PR #8, opened 2014-08-24
RxScala: Add from method to turn an Option into an Observable

jbripley:
I don't know if this is a common enough occurrence to warrant adding, but I often find myself writing this boilerplate code when jumping between collections (or other sources of optional's) and Observable code. So I thought it might be an useful addition.

The emitted Observable's are made to match the Scala behavior when doing flatMap on the result.

### DIFF
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -4571,6 +4571,22 @@ object Observable {
     toScalaObservable(rx.Observable.from(iterable.asJava, scheduler.asJavaScheduler))
   }
 
+  /**
+   * Converts an `Option` into an Observable.
+   *
+   * @param option the source `Option`
+   * @tparam T the type of item in the `Option` and the type of item
+   *           to be emitted by the resulting Observable
+   * @return an Observable that emits the item in the source `Option` or
+   *         an empty Observable if `Option` contains no item
+   */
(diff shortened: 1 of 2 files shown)


src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -4571,6 +4571,22 @@ object Observable {
...
+  /**
+   * Converts an `Option` into an Observable.
+   *
+   * @param option the source `Option`
+   * @tparam T the type of item in the `Option` and the type of item
+   *           to be emitted by the resulting Observable
+   * @return an Observable that emits the item in the source `Option` or
+   *         an empty Observable if `Option` contains no item
+   */
+  def from[T](option: Option[T]): Observable[T] = {
```
zsxwing, line 4583: `option2Iterable` can convert `Option` to `Iterable` implicitly. `from(Iterable)` should be enough. Could you give an example to explain why you need this one?

zsxwing, line 4583: I can run your unit tests without adding `from(Option)`

jbripley, line 4583: Oh, I didn't know about the implicit conversion between Option and Iterable. I'll do a PR which adds a bit to the documentation of from(Iterable) which notes this and a couple of examples to RxScalaDemo instead.

PR #9, opened 2014-08-27
Apply RxJava #1625 manually

zsxwing:
As per https://github.com/ReactiveX/RxJava/pull/1625


jbripley:
What git commands did you use for this merge? I'm trying to do the same thing for ReactiveX/RxJava#1641

zsxwing:
I used `git format-patch -1 d7895b8 --stdout > foo.patch` to create a patch of commit `d7895b8` in rxjava. Then I used `git am -p3 foo.patch` to apply the patch into rxscala. `-p3` will fix the different paths between rxjava and rxscala projects.

PR #12, opened 2014-09-12
RxScala is based on RxJava

chriswk:
Quick correction in README.md
- RxScala is based on RxJava, not RxScala


samuelgruetter:
thanks :)

PR #13, opened 2014-09-14
Remove deprecated methods and their usage

jbripley:


jmhofer:
This makes the examples compile again, therefore I merged it. Thanks!

PR #18, opened 2014-09-26
sbt build, bintray publishing, travis integration

### LINKED ISSUE #17: RxScala Build
We need to get the build and release process for RxScala functioning.

One requirement of RxScala that differs from other projects is it wants the ability to publish multiple artifacts based on different Scala versions. An issue for the Nebula plugin is at https://github.com/nebula-plugins/gradle-rxjava-project-plugin/issues/1

This is holding up progressing on RxScala as no new releases can be made beyond 0.20.x right now.

/cc @quidryan

jmhofer:
Moved build from Gradle to sbt.
Removed deprecated methods.
Added publishing to bintray.
Prepared for Travis (hopefully).
On my box and with my bintray account, this works, However, I haven't tried out the Travis script.

This is intended to fix #17.


jmhofer:
Ok, so this works with Travis. 

The only thing untested now is tagging. However, I tested the sbt command within the `publishViaTravis` script together with my own bintray repo and it worked there.

PR #23, opened 2014-10-01
Add missing Subject constructors and Subject classes from RxJava to RxScala

jbripley:
Also copies RxJava JavaDoc to ScalaDoc for each Subject as more visible documentation in Scala.


src/main/scala/rx/lang/scala/Subject.scala
```diff
@@ -29,11 +29,31 @@ trait Subject[T] extends Observable[T] with Observer[T] {
...
+  val subject = Subject[String]()
+  // observer1 will receive all onNext and onCompleted events
+  subject.subscribe(observer1)
+  subject.onNext("one")
+  subject.onNext("two")
+  // observer2 will only receive "three" and onCompleted
+  subject.subscribe(observer2)
+  subject.onNext("three")
+  subject.onCompleted()
+  </pre>
```
samuelgruetter: You should not use html tags in scaladoc, because for most of what you want to do, there's a better markdown-like version. For instance, if you use `{{{...}}}` instead of `<pre>...</pre>`, you'll get syntax highlighting. The best way to learn scaladoc is imho by example, such as by looking at any [class in the Scala standard library](https://github.com/scala/scala/blob/2.11.x/src/library/scala/collection/TraversableLike.scala).

PR #27, opened 2014-10-03
Upgrade to RxJava 1.0.0-RC5

jbripley:
- If RxJava is updated with the same signature changes with ReactiveX/RxJava#1720, the conversion from RxScala to RxJava can be simplified, without any external signature changes in RxScala.


samuelgruetter:
Nice, but let's wait with this until there's an RxJava release which contains https://github.com/ReactiveX/RxJava/pull/1720.

benjchristensen:
Are you ready to proceed with this now that 1.0.0-RC5 of RxJava is released with these changes?

jbripley:
Yes, I'll push an update to this PR later today or tomorrow.

jbripley:
Should I include the removal of groupByUntil in this PR as well, since I need to update to RxJava 1.0.0-RC5?

benjchristensen:
Good point ... yes. But the name of the pull request should be changed then so it's clear for historical purposes that it was upgrading to 1.0.0-RC5 and all related changes.

samuelgruetter:
LGTM

headinthebox:
Ship it.

PR #38, opened 2014-10-08
reasons why there is no Observable.range

samuelgruetter:
This text will finally appear in the [comparison table](http://reactivex.io/rxscala/comparison.html).


zsxwing:
LGTM. @headinthebox your opinion?

headinthebox:
I want the Scala range in RxScala. Language bindings try to be idiomatic wrt to their context.

zsxwing:
Merged as it really improves the docs. For `Observable.range`, I think we can wait until someone asks for it again.

Issue #36, opened 2014-10-08
Fail the build if CompletenessTest is not happy


samuelgruetter:
The [CompletenessTest](https://github.com/ReactiveX/RxScala/blob/0.x/src/test/scala/rx/lang/scala/CompletenessTest.scala) ensures that:
-    all methods of the Java Observable have a corresponding method in the Scala Observable, or a comment on why they don't
-    all Scala methods mentioned in the test actually exist
-    all Java methods mentioned in the test actually exist

If any of these fails, it prints some warnings. However, now that RxScala has its own build, CompletenessTest should fail the build if it isn't happy.

jbripley:
I could take a look at this one.

PR #42, opened 2014-10-09
Removes methods that also have a Scheduler parameter version by adding a default Scheduler

### LINKED ISSUE #40: Use default arguments for default schedulers
Many methods have two versions: One without scheduler, which uses some default scheduler, and one which takes a scheduler provided by the user.

Example:

``` scala
def takeRight(time: Duration): Observable[T]
def takeRight(time: Duration, scheduler: Scheduler): Observable[T]
```

Using Scala's default arguments, these two methods could be replaced by just one:

``` scala
def takeRight(time: Duration, scheduler: Scheduler = ComputationScheduler()): Observable[T]
```

For users, this _should_ not cause any breaking changes.

In addition to making the API much smaller, this also provides much better documentation: Just by looking at the signature, you know what the default scheduler is, whereas now, you have to read the documenation (which might be outdated) or even resort to reading the implementation.

This solution is less powerful than what I proposed in https://github.com/ReactiveX/RxScala/issues/16, but it's much easier to understand, and does not break any existing code. Also, since the [Default Schedulers Plugin](https://github.com/ReactiveX/RxJava/blob/master/rxjava-core/src/main/java/rx/plugins/RxJavaDefaultSchedulers.java) now provides support for custom default schedulers, there no need to solve this using implicits.

jbripley:
- Changes some methods to require parentheses (timestamp, timeInterval)
- Make CompletenessTest fail unit tests

Fixes #40, #36

### DIFF
--- a/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
+++ b/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
@@ -682,7 +682,7 @@ class RxScalaDemo extends JUnitSuite {
   }
 
   @Test def timestampExample() {
-    val timestamped = Observable.interval(100 millis).take(6).timestamp.toBlocking
+    val timestamped = Observable.interval(100 millis).take(6).timestamp().toBlocking
     for ((millis, value) <- timestamped if value > 0) {
       println(value + " at t = " + millis)
     }
@@ -1284,7 +1284,7 @@ class RxScalaDemo extends JUnitSuite {
     val o = (1 to 10).toObservable
       .zip(Observable.interval(100 millis))
       .map(_._1)
-      .timeInterval
+      .timeInterval()
     println(o.toBlocking.toList)
   }
 
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -18,6 +18,7 @@ package rx.lang.scala
 
 import rx.functions.FuncN
 import rx.lang.scala.observables.ConnectableObservable
+import rx.lang.scala.schedulers.{ImmediateScheduler, TrampolineScheduler, ComputationScheduler}
 import scala.concurrent.duration
 import java.util
 import collection.JavaConversions._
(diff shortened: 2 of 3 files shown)


jbripley:
I removed as many methods as the Scala compiler allowed me. Got this kind of error for a couple of methods (replay, repeat, tumbling, tumblingBuffer, sliding, takeRight, timeout, timer) and couldn't add a default Scheduler.

```
Error:(106, 7) in trait Observable, multiple overloaded alternatives of method tumblingBuffer define default arguments.
trait Observable[+T]
```

For `take(Duration)` I'm not sure what the problem is, but RxScalaDemo wouldn't compile without it.

samuelgruetter:
This is quite a major change, so we should carefully check that we get everything right. In order to review such a big change, it is necessary to have very clean commits, where each commit does one thing.
But here, you submit one commit which mixes many changes, such as:
-    delete methods without scheduler arguments
-    add default schedulers
-    add more documentation to some replay overloads
-    swap order of two other replay overloads
-    changes to some of the "Scheduler" sections of the methods' documentation
-    changes to CompletenessTest needed because of default schedulers
-    changes to CompletenessTest to make it fail the build if it isn't happy
-    changes to RxScalaDemo needed because of default schedulers

Could you please split them up? I'm not sure if the partition I made above is good, feel free to choose another one, but I'd like to see commits where each commit does one thing only.

Hint: You can use `git add -i` to interactively stage only parts of the changes of your working copy.

My claim that one big messy commit is not a good idea is also supported by this: You still have the following two methods:

``` scala
def throttleWithTimeout(timeout: Duration)
def throttleWithTimeout(timeout: Duration, scheduler: Scheduler = ComputationScheduler())
```

i.e. you forgot to delete the first. A more systematic approach could eliminate such mistakes.

CompletenessTest was again made an executable file. This is of course just a detail, but the fact that you missed it makes me fear that you might have missed other, more important details as well...

Moreover, it would be nice if you posted a list of all methods, indicating for each of them whether they take schedulers, and if so, if you could add a default scheduler or not.

End of my criticism ;-) Please don't take it personally. Your contributions are appreciated a lot. I'd just like to make sure we get everything right.

headinthebox:
+1. Let's try to make commits as small as possible (I am looking the mirror myself!)

jbripley:
Yes, I agree that not splitting it up was a mistake.

I think I'll split it up in these pull requests:
1. Remove non-Scheduler methods that could be removed.
2. Add default Schedulers
3. Update CompletenessTest and RxScalaDemo to compile
4. Update scala docs
5. Make CompletenessTest fail the build

Splitting up 1 and 2 will be a bit tricky since I need to know exactly which Scheduler methods supports a default scheduler, but I think I've worked that out by making this PR.

Maybe include 3 in 2 to not have failing code in-between?

samuelgruetter:
> I think I'll split it up in these pull requests

Just in case you didn't know, you can also put several commits into one pull request. That's no problem at all, because we can review each commit separately.

PR #45, opened 2014-10-12
Fail build if not all RxJava methods are mapped to a RxScala equivalent

### LINKED ISSUE #36: Fail the build if CompletenessTest is not happy

jbripley:
Fixes #36


samuelgruetter:
LGTM

PR #47, opened 2014-10-19
Add an implicit toSerialized method

zsxwing:
An convenient implicit method for `SerializedSubject`.

### DIFF
--- a/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
+++ b/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
@@ -1594,7 +1594,7 @@ class RxScalaDemo extends JUnitSuite {
   
   @Test def eventBusExample(): Unit = {
     val eventBus = Subject[String]()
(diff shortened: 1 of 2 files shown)


samuelgruetter:
Let's do the same as Java does, but that's not yet decided I guess: https://github.com/ReactiveX/RxJava/issues/1748

And btw why not an instance method instead of an extension method? The default implementation in `Subject` would be to return a new `SerializedSubject`, and `SerializedSubject` would override it and just return `this`. This would be more extensible, because users could add a `MyOwnSerializeSubject`, which does not extend `SerializedSubject`, but only `Subject`, and also returns `this` in the `toSerialized` method.

zsxwing:
IMO, an instance `toSerialized` method also cannot support a custom `MyOwnSerializeSubject` because it can not change the behaviors of other `Subject`s. That's why I prefer an extension method.

samuelgruetter:
I'm sorry, I don't get what you mean here ;-) Could you please explain?

zsxwing:
Never mind. I misunderstood your comment. Now I agree with you.

PR #52, opened 2014-11-03
Update to RxJava 1.0.0-RC9

zsxwing:
As per https://github.com/ReactiveX/RxJava/releases/tag/v1.0.0-rc.9


samuelgruetter:
LGTM
And I think that's the nicest way of solving https://github.com/ReactiveX/RxScala/issues/15 ;-)

Issue #37, opened 2014-10-08
Move RxScalaDemo & Example Coverage


samuelgruetter:
It would be nice to move the examples in [RxScalaDemo](https://github.com/ReactiveX/RxScala/blob/0.x/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala) to the [examples](https://github.com/ReactiveX/RxScala/tree/0.x/examples/src/main/scala) subproject, with one file per operator, so that it's easier to navigate through the examples, and that they can be included in the website.

Additionally, a check that each operator has an example needs to be done. This is not for aesthetic reasons or out of completeness-mania, but because throughout RxScala's history, writing examples revealed many issues with the signatures of the operators (the last being https://github.com/ReactiveX/RxScala/pull/24 which even resulted in a [signature change](https://github.com/ReactiveX/RxJava/pull/1720) in the RxJava core...)

Related to https://github.com/ReactiveX/reactivex.github.io/issues/43.

dhoepelman:
Due to project EOL status, this improvement will not be made.

PR #53, opened 2014-11-04
Check the API coverage in examples

zsxwing:
This is an initial work for #37. `APICoverage` will scan classes in the package `rx.lang.scala.examples` and report the unused APIs. It uses [BCEL](http://commons.apache.org/proper/commons-bcel/) to extract the invoked methods from byte codes.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,11 @@ name := "rxscala"
 
 lazy val root = project in file(".")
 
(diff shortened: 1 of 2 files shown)


zsxwing:
Please check the Travis CI build log here: https://travis-ci.org/ReactiveX/RxScala/builds/39962955

samuelgruetter:
Nice :) LGTM.

PR #55, opened 2014-11-06
Change signature of Observable.from(Future) to handle re-subscription

jbripley:


headinthebox:
I don't think this is a good idea for futures, you want to cache the value.

jbripley:
Caching is fine for computations, but does not work for IO calls such as HTTP requests, where retry/repeat etc are useful. Not sure how to handle both use-cases without changing the method name of the non-caching version then.

headinthebox:
Then the right thing is to make it explicit in your code by creating a new future. 
Futures are be definition hot, hence we should not repeat the computation when you subscribe again.

zsxwing:
I agree with @headinthebox 
For the following code,

``` Scala
    import scala.concurrent._
    val o = Observable.from(Future {
      println("Hello")
      1
    })
    Thread.sleep(1000)
    println("Start to subscribe")
    o.subscribe(v => println(s"receive $v"))
```

I expect it will output

```
Hello
Start to subscribe
receive 1
```

instead of

```
Start to subscribe
Hello
receive 1
```

For for IO cases, assume we have the following legacy api which is used to fetch a web page,

``` Scala
def fetchHtml(url: String): Future[String]
```

we can use `flatMap` to make it work with `retry/repeat`:

``` Scala
Observable.from(url).flatMap(url => Observable.from(fetchHtml(url))).retry/repeat
```

zsxwing:
@jbripley A git tips: Running `git config core.fileMode false` in your repo can help you avoid changing the file permission.

headinthebox:
Next time, beter to open an issue first.
Api changes need a lot of thought and care.

jbripley:
Sure, but I don't view a PR as something that is necessarily meant to be merged and more of a way to start a discussion.

I thought the updated method signature together with the included unit tests described my issue with the current signature pretty well.

@zsxwing that's an alternative, but need some additional boilerplate code to work, which is the thing Observable.from(Future) is meant to address IMO.

In our utility library at work we have a FutureObservable object with from() methods handling Twitter's Future, Scala Future and ListenableFuture. They're mostly used inside methods that all have a Observable public API and we want to be sure that a retry/repeat works regardless how the Observable was initially created or composed.

headinthebox:
BTW, I'm pretty sure you can use ()=>{ ... Conver future to observable... } to achieve you scenario in most cases. Or observable.defer, which is specifically created to make things cold.

jbripley:
If I changed the signature into an Observable.defer overload instead, would that be acceptable then?

``` scala
def defer[T](f: => Future[T])(implicit execContext: ExecutionContext): Observable[T]
```

samuelgruetter:
So if I understand correctly, this would allow us to write

``` scala
Observable.defer(createFuture())
```

instead of

``` scala
Observable.defer(Observable.from(createFuture()))
```

I think the simplification we get by this is not significant enough to justify the addition of a new method.

PR #57, opened 2014-11-08
Refactor Subscriber to support backpressure when using lift and bug fixes

zsxwing:
I found `tail` and `toXXX` didn't work after adding backpressure. To solve it, I have to change `Subscriber` to `abstract class` so that `Subscriber`s can be chained and call `request` properly.

### DIFF
--- a/src/main/scala/rx/lang/scala/Subscriber.scala
+++ b/src/main/scala/rx/lang/scala/Subscriber.scala
@@ -1,16 +1,40 @@
 package rx.lang.scala
 
-trait Subscriber[-T] extends Observer[T] with Subscription {
+abstract class Subscriber[-T](subscriber: Subscriber[_]) extends Observer[T] with Subscription {
 
   self =>
 
-  private [scala] val asJavaSubscriber: rx.Subscriber[_ >: T] = new rx.Subscriber[T] with SubscriberAdapter[T] {
-    override def onStart(): Unit = self.onStart()
-    override def onNext(value: T): Unit = self.onNext(value)
-    override def onError(error: Throwable): Unit = self.onError(error)
-    override def onCompleted(): Unit = self.onCompleted()
-    override def requestMore(n: Long): Unit = request(n)
+  def this() {
+      this(null)
   }
+
+  private[scala] val asJavaSubscriber: rx.Subscriber[_ >: T] =
+    if (subscriber == null) {
+      new rx.Subscriber[T] with SubscriberAdapter[T] {
+        override def onStart(): Unit = self.onStart()
+
+        override def onNext(value: T): Unit = self.onNext(value)
+
+        override def onError(error: Throwable): Unit = self.onError(error)
+
+        override def onCompleted(): Unit = self.onCompleted()
+
+        override def requestMore(n: Long): Unit = request(n)
+      }
+    }
+    else {
+      new rx.Subscriber[T](subscriber.asJavaSubscriber) with SubscriberAdapter[T] {
+        override def onStart(): Unit = self.onStart()
+
+        override def onNext(value: T): Unit = self.onNext(value)
+
+        override def onError(error: Throwable): Unit = self.onError(error)
+
+        override def onCompleted(): Unit = self.onCompleted()
+
+        override def requestMore(n: Long): Unit = request(n)
+      }
+    }
 
   private [scala] override val asJavaObserver: rx.Observer[_ >: T] = asJavaSubscriber
   private [scala] override val asJavaSubscription: rx.Subscription = asJavaSubscriber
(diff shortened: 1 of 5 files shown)


src/main/scala/rx/lang/scala/Subscriber.scala
```diff
@@ -74,26 +98,24 @@ object Subscriber extends ObserverFactoryMethods[Subscriber] {
...
       override def onCompleted(): Unit = c()
-    })
+    }
   }
 
   def apply[T](subscriber: Subscriber[_], onNext: T => Unit, onError: Throwable => Unit, onCompleted: () => Unit): Subscriber[T] = {
     val n = onNext; val e = onError; val c = onCompleted
-    // Java calls XXX; Scala receives XXX.
-    Subscriber(new rx.Subscriber[T](subscriber.asJavaSubscriber) {
+    new Subscriber[T](subscriber) {
```
zsxwing: @samuelgruetter could you review this PR? I don't understand why you created a new `rx.Subscriber` here. Do I miss anything?

samuelgruetter: I'm a bit confused about the RxScala/RxJava interaction in Subscriber/Observer, it doesn't look like a very systematic approach to me any more. Did you have something systematic in mind when you made this PR? Maybe you could add some docs, so that it becomes clear why it must be this way.

And regarding this `apply` signature: Why does it take a Subscriber **and** onNext/onError/onCompleted lambdas? That makes no sense to me. According to git blame, this was introduced by you in commit 
https://github.com/zsxwing/RxScala/commit/96f50d242b250949a6f03b90f426283103ed02a9

zsxwing: > I'm a bit confused about the RxScala/RxJava interaction in Subscriber/Observer, it doesn't look like a very systematic approach to me any more. Did you have something systematic in mind when you made this PR? Maybe you could add some docs, so that it becomes clear why it must be this way.

I plan to change `Observer`. However, I don't want to touch `Observer` in this PR. In your original design, why did you create `new Subscriber(rx.Subscriber(..))` instead of `new Subscriber` without a `rx.Subscriber` like this PR?

> And regarding this apply signature: Why does it take a Subscriber and onNext/onError/onCompleted lambdas? That makes no sense to me. According to git blame, this was introduced by you in commit  zsxwing@96f50d2

I added this method because it's necessary for `lift` operator. When using `lift`, we need to pass the child `Subscriber` to the new `Subscriber` to chain the `Subscription`s, e.g., the `tail` implementation:

``` scala
   def tail: Observable[T] = {
     lift {
       (subscriber: Subscriber[T]) => {
        var isFirst = true
        Subscriber[T](
          subscriber,
          (v: T) => if(isFirst) isFirst = false else subscriber.onNext(v),
          e => subscriber.onError(e),
          () => if(isFirst) subscriber.onError(new UnsupportedOperationException("tail of empty Observable")) else subscriber.onCompleted
        )
       }
     }
   }
```

However, I agree we won't need it any more since I add the constructor "Subscriber(Subscriber)" in this PR.

headinthebox: I agree, it is becoming a bit of a mess.

samuelgruetter: See also my comment on commit https://github.com/ReactiveX/RxScala/commit/fd6ec8a6b058a9ed5e90cbcb0e337f8ce023acc8.

src/main/scala/rx/lang/scala/Subscriber.scala
```diff
@@ -1,17 +1,41 @@
 package rx.lang.scala
 
-trait Subscriber[-T] extends Observer[T] with Subscription {
+abstract class Subscriber[-T](subscriber: Subscriber[_]) extends Observer[T] with Subscription {
```
samuelgruetter, line 39: So here you make a Scala Subscriber which can wrap another Scala Subscriber. Why? Is this really necessary?

zsxwing, line 39: Yes. Using `lift` to create a new operator will need this one, such as `tail`, `toMultimap`, `to[Col]`. Or a user may want to use `lift` to add his new operator.

src/main/scala/rx/lang/scala/Subscriber.scala
zsxwing: @headinthebox since you designed it, what do you think about this PR?

src/main/scala/rx/lang/scala/Subscriber.scala
samuelgruetter, line 39: "Keeping a reference to the wrapped Subscriber" is "implementing some functionality" for me, so this should be done by the RxJava core, and not be reimplemented by the RxScala adapter. I didn't try it out, but couldn't this make the design simpler?

samuelgruetter:
I like the last three commits of this PR (20f2ab6, 1c39a08, 32cbbd2), but not the first (cb7aac0), because it makes `Subscriber` even messier than it already is...

src/main/scala/rx/lang/scala/Subscriber.scala
zsxwing, line 39: What's we need to do is sending the subscriber.asJavaSubscriber to (the wrapped  Subscriber).asJavaSubscriber. We need to call the `rx.Subscriber(rx.Subscriber)` constructor: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/Subscriber.java#L47

samuelgruetter, line 39: But the `rx.Subscriber(rx.Subscriber)` constructor is protected, so that doesn't work. But can we use `rx.Subscriber.add`?

zsxwing, line 39: > But the rx.Subscriber(rx.Subscriber) constructor is protected, so that doesn't work. But can we use rx.Subscriber.add?

Subclasses can call it. I think it's better to follow the RxJava's pattern.

samuelgruetter:
If I use the `def apply[T](subscriber: rx.Subscriber[T]): Subscriber[T]` constructor and then call `request` on the resulting Scala Subscriber, I'll get a `MissingBackpressureException`, even if the underlying `rx.Subscriber` supports backpressure. Don't you think this is a problem?

zsxwing:
> If I use the def apply[T](subscriber: rx.Subscriber[T]): Subscriber[T] constructor and then call request on the resulting Scala Subscriber, I'll get a MissingBackpressureException, even if the underlying rx.Subscriber supports backpressure. Don't you think this is a problem?

Because `Subscriber.request` is `protected`, users can only call it in Subscriber or its subclasses. Therefore here the `Subscriber` returned by `apply[T](subscriber: rx.Subscriber[T])` won't call `request`.  In addition, `scala.Observable` will pass `subscriber.asJavaSubscriber` to `rx.Observable`, so if `rx.Subscriber` supports backpressure, it will still work.

samuelgruetter:
Nice explanation! I think you should add this as a comment in the source, because (at least for me) this wasn't obvious.

zsxwing:
> Nice explanation! I think you should add this as a comment in the source, because (at least for me) this wasn't obvious.

Done. I also found the correct visibility of `request` should be `protected[this]`.

samuelgruetter:
LGTM.
Sorry for the delay and the noobish remarks I made ;-) But I think it's important that the code is not only correct, but also easily convinces the readers of its correctness. Hopefully we've improved it a little bit into this direction ;-)

zsxwing:
@samuelgruetter thanks for reviewing.  Really help me improve this PR a lot. It has become nicer now.

PR #58, opened 2014-11-09
Update to RxJava 1.0.0-RC10

zsxwing:
As per https://github.com/ReactiveX/RxJava/releases/tag/v1.0.0-rc.10


zsxwing:
unnecessary

PR #67, opened 2014-11-20
Refactor toMap

zsxwing:
A PR to refactor `toMap`.

There is a breaking change in this PR: 
- Remove `def toMap[K, V] (keySelector: T => K, valueSelector: T => V, mapFactory: () => Map[K, V]): Observable[Map[K, V]]`

The other changes are still source compatible and binary compatible.

### DIFF
--- a/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
+++ b/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
@@ -993,15 +993,6 @@ class RxScalaDemo extends JUnitSuite {
     println(m.toBlocking.single)
   }
 
-  @Test def toMapExample3(): Unit = {
-    val o : Observable[String] = List("alice", "bob", "carol").toObservable
-    val keySelector = (s: String) => s.head
-    val valueSelector = (s: String) => s.tail
-    val mapFactory = () => Map(('s',"tart"))
-    val m = o.toMap(keySelector, valueSelector, mapFactory)
-    println(m.toBlocking.single)
-  }
-
   @Test def toMultimapExample1(): Unit = {
     val o : Observable[String] = List("alice", "bob", "carol", "allen", "clarke").toObservable
     val keySelector = (s: String) => s.head
--- a/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
+++ b/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
@@ -998,15 +998,7 @@ class RxScalaDemo extends JUnitSuite {
     val o : Observable[String] = List("alice", "bob", "carol").toObservable
     val keySelector = (s: String) => s.head
     val valueSelector = (s: String) => s.tail
-    val m: Observable[HashMap[Char, String]] = o.to(keySelector, valueSelector)
-    println(m.toBlocking.single)
-  }
-
-  @Test def toMapExample4(): Unit = {
-    val o : Observable[String] = List("alice", "bob", "carol").toObservable
-    val keySelector = (s: String) => s.head
-    val valueSelector = (s: String) => s.tail
-    val m = o.to[Char, String, HashMap](keySelector, valueSelector)
+    val m: Observable[HashMap[Char, String]] = o.to[HashMap, Char, String](keySelector, valueSelector)
     println(m.toBlocking.single)
   }
 
--- a/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
+++ b/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
@@ -22,6 +22,7 @@ import java.util.concurrent.TimeUnit
 import rx.lang.scala.subjects.SerializedSubject
 
 import scala.concurrent.Await
+import scala.collection.immutable.HashMap
 import scala.collection.mutable
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration.DurationInt
@@ -990,6 +991,22 @@ class RxScalaDemo extends JUnitSuite {
     val keySelector = (s: String) => s.head
     val valueSelector = (s: String) => s.tail
     val m = o.toMap(keySelector, valueSelector)
+    println(m.toBlocking.single)
+  }
+
+  @Test def toMapExample3(): Unit = {
+    val o : Observable[String] = List("alice", "bob", "carol").toObservable
+    val keySelector = (s: String) => s.head
+    val valueSelector = (s: String) => s.tail
+    val m: Observable[HashMap[Char, String]] = o.to(keySelector, valueSelector)
+    println(m.toBlocking.single)
+  }
+
+  @Test def toMapExample4(): Unit = {
+    val o : Observable[String] = List("alice", "bob", "carol").toObservable
+    val keySelector = (s: String) => s.head
+    val valueSelector = (s: String) => s.tail
+    val m = o.to[Char, String, HashMap](keySelector, valueSelector)
     println(m.toBlocking.single)
   }
 
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -3736,65 +3736,147 @@ trait Observable[+T]
   }
 
   /**
-   * Return an Observable that emits a single Map containing all items emitted by the source Observable,
-   * mapped by the keys returned by a specified `keySelector` function.
-   * <p>
-   * <img width="640" src="https://raw.github.com/wiki/Netflix/RxJava/images/rx-operators/toMap.png">
-   * <p>
-   * If more than one source item maps to the same key, the Map will contain the latest of those items.
-   *
-   * @param keySelector
-     * the function that extracts the key from a source item to be used in the Map
-   * @return an Observable that emits a single item: a Map containing the mapped items from the source
-   *         Observable
-   */
-  def toMap[K] (keySelector: T => K): Observable[Map[K, T]]= {
-    val thisJava = asJavaObservable.asInstanceOf[rx.Observable[T]]
-    val o: rx.Observable[util.Map[K, T]] = thisJava.toMap[K](keySelector)
-    toScalaObservable[util.Map[K,T]](o).map(m => m.toMap)
-  }
-
-  /**
-   * Return an Observable that emits a single Map containing values corresponding to items emitted by the
+   * Return an Observable that emits a single `Map` containing values corresponding to items emitted by the
    * source Observable, mapped by the keys returned by a specified `keySelector` function.
    * <p>
    * <img width="640" src="https://raw.github.com/wiki/Netflix/RxJava/images/rx-operators/toMap.png">
    * <p>
-   * If more than one source item maps to the same key, the Map will contain a single entry that
-   * corresponds to the latest of those items.
    *
-   * @param keySelector
-     *            the function that extracts the key from a source item to be used in the Map
-   * @param valueSelector
-     *            the function that extracts the value from a source item to be used in the Map
-   * @return an Observable that emits a single item: a HashMap containing the mapped items from the source
+   * @param keySelector the function that extracts the key from a source item to be used in the `Map`
+   * @param valueSelector the function that extracts the value from a source item to be used in the `Map`
+   * @param cbf `CanBuildFrom` to build the `Map`
+   * @return an Observable that emits a single item: a `Map` containing the mapped items from the source
    *         Observable
    */
-  def toMap[K, V] (keySelector: T => K, valueSelector: T => V) : Observable[Map[K, V]] = {
-    val thisJava = asJavaObservable.asInstanceOf[rx.Observable[T]]
-    val o: rx.Observable[util.Map[K, V]] = thisJava.toMap[K, V](keySelector, valueSelector)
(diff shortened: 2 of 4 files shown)


samuelgruetter:
I like the idea, but please give me some time to check/think about the details ;-)

Moreover, I'd like to know if anything of this PR is a breaking change. It would be nice to have a comparison of what {works/doesn't work} {with/without} this PR. Could then also be copied to the release notes of the first version containg these changes.

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -3770,31 +3768,47 @@ trait Observable[+T]
...
+  def toMutableMap[K, V >: T](keySelector: T => K): Observable[mutable.Map[K, V]] = {
+    toMap[K, V, mutable.Map](keySelector, v => v)
+  }
+
+  def toMutableMap[K, V](keySelector: T => K, valueSelector: T => V): Observable[mutable.Map[K, V]] = {
+    toMap[K, V, mutable.Map](keySelector, valueSelector)
+  }
+
+  def toTreeMap[K](keySelector: T => K)(implicit ord: Ordering[K]): Observable[scala.collection.immutable.TreeMap[K, T]] = {
+    toMap[K, T, scala.collection.immutable.TreeMap](keySelector, v => v)
```
samuelgruetter: `scala.collection.immutable.TreeMap` ---> `immutable.TreeMap`

src/test/scala/rx/lang/scala/ObservableTest.scala
```diff
@@ -428,4 +428,25 @@ class ObservableTests extends JUnitSuite {
...
+      val m2 = (0 to 100).toObservable.toMutableMap(_ % 2, _ * 100)
+      println(m2.toBlocking.single)
+    }
+    {
+      val m1 = (0 to 100).toObservable.toTreeMap(_ % 2)
+      println(m1.toBlocking.single)
+      val m2 = (0 to 100).toObservable.toTreeMap(_ % 2, _ * 100)
+      println(m2.toBlocking.single)
+    }
+  }
```
samuelgruetter, line 444: will need more tests with assertions, and one def per test

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -3770,31 +3768,47 @@ trait Observable[+T]
    * @return an Observable that emits a single item: a HashMap containing the mapped items from the source
    *         Observable
    */
-  def toMap[K, V] (keySelector: T => K, valueSelector: T => V) : Observable[Map[K, V]] = {
-    val thisJava = asJavaObservable.asInstanceOf[rx.Observable[T]]
-    val o: rx.Observable[util.Map[K, V]] = thisJava.toMap[K, V](keySelector, valueSelector)
-    toScalaObservable[util.Map[K, V]](o).map(m => m.toMap)
+  def toMap[K, V, M[_, _]](keySelector: T => K, valueSelector: T => V)(implicit cbf: CanBuildFrom[Nothing, (K, V), M[K, V]] = Map.canBuildFrom[K, V]): Observable[M[K, V]] = {
```
samuelgruetter: You should explain how contravariance makes `CanBuildFrom[Nothing, ...]` work. Moreover, having a default value for an implicit seems a bit dangerous at first sight. If there's an implicit in scope, it takes precedence over the default value. But what if there are several matching implicits? Do I get an "ambiguous implicits" error or does it fall back to the default value? Can I find this somewhere in the Scala spec? What relevant implicits are typically in scope when this function is called? Did you invent this pattern of combining an implicit `CanBuildFrom[Nothing, ...]` with a default argument yourself, or are there other libraries which do this as well?

samuelgruetter: Plenty of questions when reading this code... And I guess you've already asked and answered most of them yourself while writing this code, so could you please share your insights with us? That's what source code comments or commit messages are good for ;-)

zsxwing:
I updated the description and override the PR with a formal commit. Sorry that I should have done it at first.

zsxwing:
> having a default value for an implicit seems a bit dangerous at first sight.

You're right. Thanks for pointing out it. I created a helper method 

``` Scala
def to[K, V, M[_, _]](keySelector: T => K, valueSelector: T => V)(implicit cbf: CanBuildFrom[Nothing, (K, V), M[K, V]]): Observable[M[K, V]] = {
```

and used to implement `toMap`. This will make `toXXXMap` be consistent with `toList`, `toSeq`.

samuelgruetter:
Btw did you try out if you can implement `toMap` just as

```
this.map(elem => (keySelector(elem), valueSelector(elem))).to[Map]
```

zsxwing:
It won't work. Here is the signature of `to`:

``` Scala
def to[Col[_]](implicit cbf: CanBuildFrom[Nothing, T, Col[T @uncheckedVariance]]): Observable[Col[T @uncheckedVariance]]
```

It expects a `Col` with one type parameter and will return `Col[T]`. 
`this.map(elem => (keySelector(elem), valueSelector(elem))).to[Map]` won't work because:
- Map has two type parameters
- toMap should return `Observable[Map[K, V]]` instead of `Observable[Map[(K, V)]]`

samuelgruetter:
I see.
Something else: Let's add the `toMap` of Scala Lists:

``` scala
def toMap[T, U](implicit ev: A <:< (T, U)): Map[T, U] 
```

examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
```diff
@@ -993,15 +993,6 @@ class RxScalaDemo extends JUnitSuite {
...
 
-  @Test def toMapExample3(): Unit = {
-    val o : Observable[String] = List("alice", "bob", "carol").toObservable
-    val keySelector = (s: String) => s.head
-    val valueSelector = (s: String) => s.tail
-    val mapFactory = () => Map(('s',"tart"))
-    val m = o.toMap(keySelector, valueSelector, mapFactory)
-    println(m.toBlocking.single)
-  }
-
```
samuelgruetter, line 1004: If you remove this example, you should show how we can do this now. Not necessarily with a non-empty starting map, but maybe with a HashMap from the collection library?

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -3736,65 +3736,147 @@ trait Observable[+T]
...
+   * corresponds to the latest of those items.
+   *
+   * @param keySelector the function that extracts the key from a source item to be used in the `immutable.TreeMap`
+   * @param valueSelector the function that extracts the value from a source item to be used in the `immutable.TreeMap`
+   * @param ord the implicit ordering used to compare the keys.
+   * @return an Observable that emits a single item: an `immutable.TreeMap` containing the mapped items from the source
+   *         Observable
+   */
+  def toTreeMap[K, V](keySelector: T => K, valueSelector: T => V)(implicit ord: Ordering[K]): Observable[immutable.TreeMap[K, V]] = {
+    to[K, V, immutable.TreeMap](keySelector, valueSelector)
```
samuelgruetter: I'm not sure if we should have these `toXxxMap` methods. Now we just have `TreeMap` and `MutableMap`, but to cover all use cases, we would need many more. I'd find it nicer to have a light-weight way for users to specify the map type they want.
Maybe something like `myObservable.to[immutable.TreeMap]`, but probably we can't overload the `to` method like this...

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -3736,65 +3736,147 @@ trait Observable[+T]
...
 
   /**
-   * Return an Observable that emits a single Map containing values corresponding to items emitted by the
+   * Return an Observable that emits a single `immutable.Map` containing values corresponding to items emitted by the
    * source Observable, mapped by the keys returned by a specified `keySelector` function.
    * <p>
    * <img width="640" src="https://raw.github.com/wiki/Netflix/RxJava/images/rx-operators/toMap.png">
    * <p>
-   * If more than one source item maps to the same key, the Map will contain a single entry that
+   * If more than one source item maps to the same key, the `immutable.Map` will contain a single entry that
```
samuelgruetter: No need to clutter the docs with `immutable.` prefixes ;-) Unless otherwise stated, everyone will assume it's immutable.

zsxwing:
> Let's add the toMap of Scala Lists:

Done

> No need to clutter the docs with immutable. prefixes 

Done

> If you remove this example, you should show how we can do this now.

Added two examples to show how to use `to[K, V, M[_, _]]`. However, I cannot find a way to make the compiler infer the `K, V` types.

> I'm not sure if we should have these toXxxMap methods.

Maybe it's better not to add `toXxxMap` methods in this PR?

samuelgruetter:
> Maybe it's better not to add toXxxMap methods in this PR?

Yes, I think so.

examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
```diff
@@ -997,8 +998,15 @@ class RxScalaDemo extends JUnitSuite {
...
-    val m = o.toMap(keySelector, valueSelector, mapFactory)
+    val m: Observable[HashMap[Char, String]] = o.to(keySelector, valueSelector)
+    println(m.toBlocking.single)
+  }
+
+  @Test def toMapExample4(): Unit = {
+    val o : Observable[String] = List("alice", "bob", "carol").toObservable
+    val keySelector = (s: String) => s.head
+    val valueSelector = (s: String) => s.tail
+    val m = o.to[Char, String, HashMap](keySelector, valueSelector)
```
samuelgruetter: I don't quite like these two examples because none of them shows explicitly what's happening. I'd only make one example, but add all types there:

``` scala
    val m: Observable[HashMap[Char, String]] = o.to[Char, String, HashMap](keySelector, valueSelector)
```

IMHO that's more readable.

samuelgruetter: And while I'm reading this, I wonder why we don't swap the order of the type params:

``` scala
o.to[HashMap, Char, String](keySelector, valueSelector)
```

corresponds better to the order of pronouncing what we're doing: we say "convert o to a HashMap of Char/String".

samuelgruetter:
We should not forget to make `toMultimap` consistent with `toMap`, but we can also do that in a separate PR.

zsxwing:
> swap the order of the type params

Sounds great. Done.

zsxwing:
I cannot figure out how to refactor `toMultimap` now. Here is my attempt. However, the problem is `how to add value to b?`

``` Scala
  def to[M[_, _], Col[_], K, V](keySelector: T => K, valueSelector: T => V)
                               (implicit mcbf: CanBuildFrom[Nothing, (K, Col[V]), M[K, Col[V]]], 
                                cbf: CanBuildFrom[Nothing, V, Col[V]]): Observable[M[K, Col[V]]] = {
    lift {
      (subscriber: Subscriber[M[K, Col[V]]]) => {
        new Subscriber[T](subscriber) {
          val b = mcbf()

          override def onStart(): Unit = request(Long.MaxValue)

          override def onNext(t: T): Unit = {
            val key = keySelector(t)
            val value = valueSelector(t)
            // how to add value to b?
            // b expects (K, Col[V])
          }

          override def onError(e: Throwable): Unit = {
            subscriber.onError(e)
          }

          override def onCompleted(): Unit = {
            subscriber.onNext(b.result)
            subscriber.onCompleted()
          }
        }
      }
    }
  }

```

samuelgruetter:
We could also use `scala.collection.mutable.MultiMap`, which has an `addBinding` method. Since it's mutable, there's no need to use a builder, we could just start with an empty mutable multimap and add one key/value pair after the other using `addBinding`. The question is, however, to what extent we can/want to abstract over the Map type and Set type used by MultiMap.

zsxwing:
Sent another PR #68 for `toMultimap` as per your suggestion.

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -3736,65 +3736,90 @@ trait Observable[+T]
...
-  def toMap[K, V] (keySelector: T => K, valueSelector: T => V, mapFactory: () => Map[K, V]): Observable[Map[K, V]] = {
-    val thisJava = asJavaObservable.asInstanceOf[rx.Observable[T]]
-    val o: rx.Observable[util.Map[K, V]] = thisJava.toMap[K, V](keySelector, valueSelector)
-    toScalaObservable[util.Map[K, V]](o).map(m => mapFactory() ++ m.toMap)
+   * @param keySelector the function that extracts the key from a source item to be used in the `Map`
+   * @param valueSelector the function that extracts the value from a source item to be used in the `Map`
+   * @return an Observable that emits a single item: an `Map` containing the mapped items from the source
+   *         Observable
+   */
+  def toMap[K, V](keySelector: T => K, valueSelector: T => V): Observable[Map[K, V]] = {
```
samuelgruetter, line 3821: Did you try if you can define a default argument of `v => v` for valueSelector? Not sure if this can work because scalac would have to infer from this that T must be the same as V.

zsxwing, line 3821: not work...

zsxwing:
@samuelgruetter further suggestion?

samuelgruetter:
LGTM

PR #68, opened 2014-11-28
Refactor toMultimap

zsxwing:
This PR includes the following breaking changes:
- Change the return type from `Observable[scala.collection.Map[K, Seq[T]]]` to `Observable[mutable.MultiMap[K, V]]`.
- Change the method name `toMultimap` to `toMultiMap` to make it consistent to the return type.
- Remove `toMultimap(keySelector, valueSelector, mapFactory, bufferFactory)`. You can override `MultiMap.makeSet` to create your custom bufferFactory. E.g.,

``` Scala
     val o : Observable[String] = List("alice", "bob", "carol", "allen", "clarke").toObservable
     val keySelector = (s: String) => s.head
     val valueSelector = (s: String) => s.tail
     val mapFactory = () => new mutable.HashMap[Char, mutable.Set[String]] with mutable.MultiMap[Char, String] {
       override def makeSet = new mutable.TreeSet[String]
     }
     val m = o.toMultiMap(keySelector, valueSelector, mapFactory)
     println(m.toBlocking.single)
```

### DIFF
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -3960,84 +3960,60 @@ trait Observable[+T]
   }
 
   /**
-   * Returns an Observable that emits a single `Map` that contains an `Seq` of items emitted by the
-   * source Observable keyed by a specified `keySelector` function.
+   * Returns an Observable that emits a single `mutable.MultiMap` that contains items emitted by the
+   * source Observable keyed by a specified `keySelector` function. The items having the same
+   * key will be put into a `Set`.
    *
    * <img width="640" height="305" src="https://raw.github.com/wiki/Netflix/RxJava/images/rx-operators/toMultiMap.png">
    *
-   * @param keySelector the function that extracts the key from the source items to be used as key in the HashMap
-   * @return an Observable that emits a single item: a `Map` that contains an `Seq` of items mapped from
-   *         the source Observable
+   * @param keySelector the function that extracts the key from the source items to be used as key in the `mutable.MultiMap`
+   * @return an Observable that emits a single item: a `mutable.MultiMap` that contains items emitted by the
+   *         source Observable keyed by a specified `keySelector` function.
    */
-  def toMultimap[K](keySelector: T => K): Observable[scala.collection.Map[K, Seq[T]]] = {
-    toMultimap(keySelector, k => k)
+  def toMultiMap[K, V >: T](keySelector: T => K): Observable[mutable.MultiMap[K, V]] = {
+    toMultiMap(keySelector, k => k)
   }
 
   /**
-   * Returns an Observable that emits a single `Map` that contains an `Seq` of values extracted by a
+   * Returns an Observable that emits a single `mutable.MultiMap` that contains values extracted by a
    * specified `valueSelector` function from items emitted by the source Observable, keyed by a
-   * specified `keySelector` function.
+   * specified `keySelector` function. The values having the same key will be put into a `Set`.
    *
    * <img width="640" height="305" src="https://raw.github.com/wiki/Netflix/RxJava/images/rx-operators/toMultiMap.png">
    *
-   * @param keySelector the function that extracts a key from the source items to be used as key in the HashMap
-   * @param valueSelector the function that extracts a value from the source items to be used as value in the HashMap
-   * @return an Observable that emits a single item: a `Map` that contains an `Seq` of items mapped from
+   * @param keySelector the function that extracts a key from the source items to be used as key in the `mutable.MultiMap`
+   * @param valueSelector the function that extracts a value from the source items to be used as value in the `mutable.MultiMap`
+   * @return an Observable that emits a single item: a `mutable.MultiMap` that contains keys and values mapped from
    *         the source Observable
    */
-  def toMultimap[K, V](keySelector: T => K, valueSelector: T => V): Observable[scala.collection.Map[K, Seq[V]]] = {
-    toMultimap(keySelector, valueSelector, () => mutable.Map[K, mutable.Buffer[V]]())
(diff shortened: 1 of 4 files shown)


src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -3960,84 +3960,60 @@ trait Observable[+T]
...
+   * @param keySelector the function that extracts a key from the source items to be used as the key in the `mutable.MultiMap`
+   * @param valueSelector the function that extracts a value from the source items to be used as the value in the `mutable.MultiMap`
+   * @param multiMapFactory the function that returns a `mutable.MultiMap` instance to be used
+   * @return an Observable that emits a single item: a `mutable.MultiMap` that contains keys and values mapped from the source Observable.
    */
-  def toMultimap[K, V, B <: mutable.Buffer[V], M <: mutable.Map[K, B]](keySelector: T => K, valueSelector: T => V, mapFactory: () => M, bufferFactory: K => B): Observable[M] = {
-    // It's complicated to convert `mutable.Map[K, mutable.Buffer[V]]` to `java.util.Map[K, java.util.Collection[V]]`,
-    // so RxScala implements `toMultimap` directly.
-    // Choosing `mutable.Buffer/Map` is because `append/update` is necessary to implement an efficient `toMultimap`.
+  def toMultiMap[K, V, M <: mutable.MultiMap[K, V]](keySelector: T => K, valueSelector: T => V, multiMapFactory: () => M): Observable[M] = {
```
samuelgruetter: -    I think a by name parameter (i.e. `multiMapFactory: => M`) would be more scala-ish than `multiMapFactory: () => M`.
-    Can you add default arguments for `multiMapFactory` and maybe even for `valueSelector`? Then you could remove some overloads and get a smaller API.

zsxwing: Default value for `multiMapFactory` does not work. If I change it to

``` Scala
  def toMultiMap[K, V, M <: mutable.MultiMap[K, V]](
       keySelector: T => K,
       valueSelector: T => V,
       multiMapFactory: => M = new mutable.HashMap[K, mutable.Set[V]] with mutable.MultiMap[K, V]): Observable[M] = {
```

There will be some error such as

```
RxScala/src/test/scala/rx/lang/scala/ObservableTest.scala:242: type mismatch;
[error]  found   : String => String
[error]  required: String => V
[error]     val o = Observable.just("a", "b", "cc", "dd").toMultiMap(_.length, s => s + s)
```

Looks the compiler can not infer the type `V` from `_.length`, `s => s + s` and the default value `new mutable.HashMap[String, mutable.Set[V]] with mutable.MultiMap[String, V]`.

zsxwing:
When I saw the following signature, 

``` Scala
toMultiMap[K, V, M <: mutable.MultiMap[K, V]](keySelector: T => K, valueSelector: T => V, multiMapFactory: => M)
```

I thought it might be better to use currying for `multiMapFactory`.
So I tried the following two ways:
1.

``` Scala
toMultiMap[K, V, M <: mutable.MultiMap[K, V]](multiMapFactory: => M)(keySelector: T => K, valueSelector: T => V)
```

It does not work because

```
RxScala/src/main/scala/rx/lang/scala/Observable.scala:3990: overloaded method value toMultiMap with alternatives:
[error]   [K, V, M <: scala.collection.mutable.MultiMap[K,V]](multiMapFactory: => M)(keySelector: T => K, valueSelector: T => V)rx.lang.scala.Observable[M] <and>
[error]   [K, V >: T](keySelector: T => K)rx.lang.scala.Observable[scala.collection.mutable.MultiMap[K,V]]
[error]  cannot be applied to (scala.collection.mutable.HashMap[K,scala.collection.mutable.Set[V]] with scala.collection.mutable.MultiMap[K,V])
[error]     toMultiMap(new mutable.HashMap[K, mutable.Set[V]] with mutable.MultiMap[K, V])(keySelector, valueSelector)
[error]     ^

```

2.

``` Scala
toMultiMap[K, V, M <: mutable.MultiMap[K, V]](keySelector: T => K, valueSelector: T => V)(multiMapFactory: => M)
```

It also does not work because

```
RxScala/src/main/scala/rx/lang/scala/Observable.scala:3990: ambiguous reference to overloaded definition,
[error] both method toMultiMap in trait Observable of type [K, V, M <: scala.collection.mutable.MultiMap[K,V]](keySelector: T => K, valueSelector: T => V)(multiMapFactory: => M)rx.lang.scala.Observable[M]
[error] and  method toMultiMap in trait Observable of type [K, V](keySelector: T => K, valueSelector: T => V)rx.lang.scala.Observable[scala.collection.mutable.MultiMap[K,V]]
[error] match argument types (T => K,T => V)
[error]     toMultiMap(keySelector, valueSelector)(new mutable.HashMap[K, mutable.Set[V]] with mutable.MultiMap[K, V])
[error]     ^
```

At last, looks the following signature is best now:

``` Scala
toMultiMap[K, V, M <: mutable.MultiMap[K, V]](keySelector: T => K, valueSelector: T => V, multiMapFactory: => M)
```

zsxwing:
@samuelgruetter further suggestion?

samuelgruetter:
LGTM.
Too bad we can't have the curried version, but I didn't find anything either.
And I think it's acceptable that in `toMap`, we use `CanBuildFrom`, whereas in `toMultiMap`, we use a `mapFactory`, because Scala's `MultiMap` was not designed to work with `CanBuildFrom`, so it's not "our fault".

samuelgruetter:
@zsxwing could you please write some release notes giving a summary about the changes of `toMap` and `toMultiMap`? So that users know how to update their code, and why certain methods were removed.

zsxwing:
> @zsxwing could you please write some release notes giving a summary about the changes of toMap and toMultiMap? So that users know how to update their code, and why certain methods were removed.

OK.

PR #75, opened 2014-12-07
Remove ReleaseNotes.md

zsxwing:
ReleaseNotes.md is out of data. It's for 0.16.0. We should remove it. For new release notes, we should add them to `CHANGES.md`, just like RxJava.


samuelgruetter:
LGTM

Issue #79, opened 2014-12-07
onErrorResumeNext type inference annoyance


jbripley:
Since there are two overloads for `onErrorResumeNext`, it is not possible to use the common Scala idiom of skipping the match statement for a partial function sent in to `onErrorResumeNext(Throwable => Observable[U])`:

Ex. A

``` Scala
o.onErrorResumeNext({
  case io:IOException => Observable.just(20, 30, 40)
  case e:Throwable => Observable.just(10, 11, 12)
})
```

Instead it needs to be written with the more verbose version, like this:

Ex. B

``` Scala
o.onErrorResumeNext({ error =>
  error match {
    case io: IOException => Observable.just(20, 30, 40)
    case e: Throwable => Observable.just(10, 11, 12)
  }
})
```

I'm not the most familiar with the Scala type system, but would it be possible to add more to the type definition of either `onErrorResumeNext` methods to allow the compiler to allow Ex. A to work? If not, would it be possible to rename `onErrorResumeNext(Throwable => Observable[U])` instead? Maybe to something like `onErrorResumeWith` or `onErrorResumeUsing`.

samuelgruetter:
We're not the first to have this problem: http://scala-programming-language.1934581.n4.nabble.com/Type-inference-craps-out-tt2134282.html#a2134476
Very often, Scala's features stop working nicely when combined with overloading, and noone invests time on fixing it, because everyone says "overloading is bad"...

The simplest solution to this would be to remove the `onErrorResumeNext[U >: T](Observable[U])` overload, because it can easily be obtained with `o.onErrorResumeNext(_ => resumeSeq)`. What do you think?

jbripley:
Yes, removing `onErrorResumeNext[U >: T](Observable[U])` sounds like the most pragmatic way to solve this. I'll make PR using that solution and we'll see if it runs into any problems.

zsxwing:
Fixed in #82

PR #82, opened 2014-12-07
Fix onErrorResumeNext partial function ambiguity problem

jbripley:
As described in #79, the current implementation of having two overloaded onErrorResumeNext methods leads to ambiguity problems with the common Scala partial function shorthand idiom. This pull request solves this by removing one of the onErrorResumeNext methods, since the functionality of that method can be accomplished by using the remaining method.


zsxwing:
LGTM. Hold off until 0.24.0.

Issue #80, opened 2014-12-07
Observable error


headinthebox:
```
def error[T](exception: Throwable): Observable[T] = {
    toScalaObservable[T](rx.Observable.error(exception))
  }
```

This can return `Observbale[Nothing]` like `empty` and `never` or not?

samuelgruetter:
Yes it can, because if no type parameter for `T` is given, the compiler infers the most precise type possible, which is `Nothing`:

``` scala
val o1 = Observable.error(new Exception())      // creates an Observable[Nothing]
```

samuelgruetter:
Oops, looks like I misunderstood the question, and then you misunderstood my answer.

I understood the question as "Can we use `Observable.error` _as if_ it was defined without a type parameter and returning `Observable[Nothing]`"? And the answer to this question is yes.

But I would keep the additional type parameter in the current signature, because it has the advantage that we can give the desired type parameter explicitly if we wish to do so, which often improves type inference, as @zsxwing's example in https://github.com/ReactiveX/RxScala/pull/83 shows.

zsxwing:
I think `empty`, `never` and `error` have the same problem: should we use `Observable[Nothing]`?

I tend to `yes` and prefer to make these methods consistent because:
- It's much like `Nil`. `Nil` also has similar problem. E.g.,

``` Scala
val o: List[Int] = Nil.map(v => v)
val x = 1
println(x + (Nil.head: Int))
val y: Int = Nil.head
```
- My examples in #83 are contrived. I don't think people will write such codes. When they use `empty`, `never` or `error`, they usually have some other Observable with a known type parameter in hand and combining these `Observable`s will work well with type inference.

samuelgruetter:
I agree that we should make `empty`, `never` and `error` consistant. But I'd prefer adding type params to all of them because:

1) Scala's `List.empty` does so as well:

``` scala
def empty[A]: List[A] 
```

2) There's no program which stops working if we add type parameters, but some stop working if we remove them.

3) This is a chance to help users when Scala's type inference is not smart enough.

zsxwing:
Great points. I'm convinced.

headinthebox:
Disagree. I absolutely hate to see any types. It is super irritating to be forced to specify the type in error. We should leverage covariance.

As long as we are not at 1.0 we _should_ break people as we iterate towards a stable API.

**I'll veto the change to add type parameters.**

zsxwing:
@headinthebox 

[...]

PR #83, opened 2014-12-08
Optimize the 'error' signature

zsxwing:
As per #80

This will break the source compatibility. E.g., `Observable.error[Int](new Exception(msg)).firstOrElse(10).toBlocking.single` won't be compiled.

Hold off until we prepare to release 0.24.0.

### DIFF
--- a/src/test/scala/rx/lang/scala/ObservableTest.scala
+++ b/src/test/scala/rx/lang/scala/ObservableTest.scala
@@ -94,7 +94,7 @@ class ObservableTests extends JUnitSuite {
     val msg = "msg6251"
     var receivedMsg = "none"
     try {
-      Observable.error[Int](new Exception(msg)).firstOrElse(10).toBlocking.single
+      Observable.error(new Exception(msg)).firstOrElse(10).toBlocking.single
     } catch {
       case e: Exception => receivedMsg = e.getCause().getMessage()
     }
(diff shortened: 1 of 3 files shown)


samuelgruetter:
I would not make this change for the reasons I explained in https://github.com/ReactiveX/RxScala/issues/80#issuecomment-66110613

src/test/scala/rx/lang/scala/ObservableTest.scala
```diff
@@ -428,4 +428,15 @@ class ObservableTests extends JUnitSuite {
     assertEquals(expectedMap3, m3.toBlocking.single)
   }
 
+  @Ignore // Don't run it. Only test if it can be compiled
+  def testErrorSignature() {
+    val o: Observable[Int] = Observable.error(new RuntimeException("Oops")).map(v => v)
+    println(o.toBlocking.single)
+
+    val x = 1
+    println(x + (Observable.error(new RuntimeException("Oops")).toBlocking.single: Int))
```
zsxwing, line 437: @samuelgruetter Just revisited this one and found `error[T]` also needs to specify `: Int`. Now I cannot find a case that `error[T]` is better than `error[Nothing]`. So how about changing to `error[Nothing]`?

headinthebox, line 437: Yes !

samuelgruetter, line 437: Some people might prefer `Observable.error[Int](...)` over a `... : Int` ascription... ;-)
But I guess such examples will be very rare, only occur in constructed examples, so I'd be ok with `def error(exception: Throwable): Observable[Nothing]` (but note that it's a breaking change).

PR #86, opened 2014-12-08
Fix issue #85 that Subscription.isUnsubscribed returns a wrong value

### LINKED ISSUE #85: Auto unsubscribe bug
The following should print `true` and `true` (does in Java for apply).

``` scala
object Main {
  def main(args: Array[String]): Unit = {
    val xs = Observable.create[Nothing](o => {
      o.onCompleted()
      Subscription{ println("...") }
    })
    val s = xs.subscribe()
    println(s.isUnsubscribed)
    val ys = Observable[Nothing](o => {
      o.onCompleted()
    })
    val z = ys.subscribe()
    println(z.isUnsubscribed)
    readLine()
  }
}
```

zsxwing:
Fix #85

### DIFF
--- a/src/main/scala/rx/lang/scala/Subscription.scala
+++ b/src/main/scala/rx/lang/scala/Subscription.scala
@@ -32,7 +32,7 @@
 
 package rx.lang.scala
 
-import java.util.concurrent.atomic.AtomicBoolean
+import rx.functions.Action0
 
 /**
  * Subscriptions are returned from all `Observable.subscribe` methods to allow unsubscribing.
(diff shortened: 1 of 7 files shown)


src/main/scala/rx/lang/scala/Subscription.scala
```diff
@@ -41,23 +41,22 @@ import java.util.concurrent.atomic.AtomicBoolean
  */
 trait Subscription {
 
-  private [scala] val unsubscribed = new AtomicBoolean(false)
-  private [scala] val asJavaSubscription: rx.Subscription = new rx.Subscription {
-    override def unsubscribe() { unsubscribed.compareAndSet(false, true) }
-    override def isUnsubscribed(): Boolean = { unsubscribed.get() }
-  }
-
+  private[scala] val asJavaSubscription: rx.Subscription = rx.subscriptions.BooleanSubscription.create()
```
samuelgruetter, line 44: I like this change: We're not reimplementing the wheel any more, and we follow the same pattern as Subscriber (default asJava val, which is overridden in apply constructors if needed).

samuelgruetter:
LGTM

PR #87, opened 2014-12-08
Add a type parameter for 'empty' and 'never'

zsxwing:
As per #80. 

~~Adding type parameters breaks the source compatibility. Hold off until we prepare to release 0.24.0.~~


samuelgruetter:
> Adding type parameters breaks the source compatibility.

Why? Can you give an example which works before this PR but not after?

zsxwing:
> Why? Can you give an example which works before this PR but not after?

Never mind. I was wrong...

headinthebox:
No ... it should be `Nothing`!

headinthebox:
I'm going to close this; let's not make our lives worse as a response to making our lives easier.

Issue #77, opened 2014-12-07
rx.schedulers.Schedulers.from(Executor)


samuelgruetter:
Should we add a Scala equivalent for `rx.schedulers.Schedulers.from(Executor)`? Would it take a `scala.concurrent.ExecutionContext`? And what's the relation/difference between a Java `Executor` and a Scala `ExecutionContext`?

headinthebox:
>  And what's the relation/difference between a Java Executor and a Scala ExecutionContext?

I always wondered about that; why did Scala no reuse the standard Java executors?

jbripley:
I think we should add it. It would allow someone to run Scala Futures and Observables against the same underlying thread pool, which I imagine could be useful.

zsxwing:
> And what's the relation/difference between a Java Executor and a Scala ExecutionContext?
> And somehow, Java and Scala are compatible, can we exploit this better? 

Scala `ExecutionContextImpl` uses `Java Executor`: https://github.com/scala/scala/blob/2.11.x/src/library/scala/concurrent/impl/ExecutionContextImpl.scala#L27 but we cannot get it via any public API, that's why I created a `Java Executor` wrapper for `Scala ExecutionContext`.

> And do we also have to care about ExecutorServices?

Because Scala doesn't have something like `Java ScheduledExecutorService`, we can ignore `ScheduledExecutorService`. Therefore in RxJava, `ExecutorScheduler` only uses `Executor.execute`: https://github.com/ReactiveX/RxJava/blob/fa927ab79f1996e7b13caa3b79d7f96be86856d4/src/main/java/rx/schedulers/ExecutorScheduler.java#L81 ,so we don't need to care about `ExecutorService`.

zsxwing:
Fixed in #91

PR #91, opened 2014-12-09
Add ExecutionContextScheduler for Scala ExecutionContext

zsxwing:
As per #77

### DIFF
--- a/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
+++ b/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala
@@ -1360,26 +1360,23 @@ class RxScalaDemo extends JUnitSuite {
     subscription.unsubscribe()
   }
 
-  @Test def executionContextSchedulerExample(): Unit = {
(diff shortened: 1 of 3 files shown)


samuelgruetter:
Are you sure that this is "the right way" to do it? Can you answer the question I asked in https://github.com/ReactiveX/RxScala/issues/77#issue-51220564? And somehow, Java and Scala are compatible, can we exploit this better? The following compiles:

``` scala
val ec: java.util.concurrent.Executor = scala.concurrent.ExecutionContext.Implicits.global
```

And do we also have to care about `ExecutorService`s?
I'd prefer adding nothing than something we don't understand...

zsxwing:
Answered in #77

zsxwing:
@samuelgruetter could you revisit this PR? I added unit tests to verify the thread name. For `ExecutionContextExecutor`, since it's already a `java.util.concurrent.Executor`, we don't need to wrap it.

zsxwing:
Updated the examples.

samuelgruetter:
LGTM

PR #97, opened 2014-12-15
Add the version table

zsxwing:
The table shows the compatible RxJava version for RxScala.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,15 @@ Since RxScala is part of the RxJava family the communication channels are simila
 
 ## Versioning
 
-Before 0.21, RxScala 0.x was based on RxJava 0.x.
-From 0.21, RxScala is based on RxJava 1.0.
+| RxScala version | Compatible RxJava version |
+| ------------------- | ------------------------- |
+| 0.23.*<sup>[1]</sup> | 1.0.* |
+| 0.22.0 | 1.0.0-rc.5 |
+| 0.21.1 | 1.0.0-rc.3 |
+| 0.X.Y (X < 21)<sup>[2]</sup> | 0.X.Y |
 
-As of 1.0.0 semantic versioning will be used.
+[1] You can use any release of RxScala 0.23 with any release of RxJava 1.0. E.g, use RxScala 0.23.0 with RxJava 1.0.1 <br/>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,10 +79,15 @@ Since RxScala is part of the RxJava family the communication channels are simila
...
-
-As of 1.0.0 semantic versioning will be used.
+| RxScala version | Compatible RxJava version |
+| ------------------- | ------------------------- |
+| 0.23.*<sup>[1]</sup> | 1.0.* |
+| 0.22.0 | 1.0.0-rc.5 |
+| 0.21.1 | 1.0.0-rc.3 |
+| 0.X.Y (X < 21)<sup>[2]</sup> | 0.X.Y |
+
+[1] You can use any release of RxScala 0.23 with any release of RxJava 1.0. E.g, use RxScala 0.23.0 with RxJava 1.0.1 <br/>
```
samuelgruetter, line 89: Are you sure? In the POM at http://search.maven.org/#artifactdetails|io.reactivex|rxscala_2.11|0.23.0|jar, I read

```
    <dependency>
        <groupId>io.reactivex</groupId>
        <artifactId>rxjava</artifactId>
        <version>1.0.2</version>
    </dependency>
```

so I guess it will only work with `1.0.2`...

zsxwing, line 89: Here. I mean people can specify other RxJava version in their pom.xml, build.gradle, or build.sbt, e.g.,

```
libraryDependencies ++= Seq(
  "io.reactivex" % "rxjava" % "1.0.3",
  "io.reactivex" %% "rxscala" % "0.23.0")
```

in case RxScala doesn't release a version for RxJava 1.0.3 in time, but they want to update RxJava to 1.0.3 for some critical bug fixes.

zsxwing, line 89: So people can know the compatible version range of RxJava from this table and update RxJava if they need.

samuelgruetter, line 89: Right. So I tried it out and sbt is smarter than I thought: If a library is required multiple times, but different versions, sbt doesn't complain, nor put multiple versions of the same library on the classpath, but it just picks the most recent version of all mentioned versions.

samuelgruetter:
LGTM

PR #103, opened 2014-12-22
Add a release note for 0.23.1

zsxwing:


samuelgruetter:
Just a detail: RxJava includes the date of each release in the titles: https://github.com/ReactiveX/RxJava/blob/1.x/CHANGES.md
Would be nice to do so in RxScala as well.

zsxwing:
For 0.23.1, we won't know the date until we release it. So I will add the data when I release 0.23.1. The date is in the PST(UTC-8) time zone.

samuelgruetter:
:+1:

PR #108, opened 2014-12-28
Improve docs

zsxwing:
1. Add more docs for Subscriber
2. Fix warning of `sbt doc`
3. CRLF => LF in SerializedSubject.scala


zsxwing:
@samuelgruetter thanks you for the good catches. They are unintentional.

PR #122, opened 2015-01-23
Update the `subscribe` docs

### LINKED ISSUE #121: ScalaDoc for Observable.subscribe() seems to be incorrect.
http://reactivex.io/rxscala/scaladoc/index.html#rx.lang.scala.Observable@subscribe():rx.lang.scala.Subscription

subscribe() takes no parameters, so the following documentation seems to be incorrect "Call this method to subscribe an rx.lang.scala.Observer for receiving items and notifications from the Observable."

zsxwing:
Fix #121

### DIFF
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -16,6 +16,7 @@
 
 package rx.lang.scala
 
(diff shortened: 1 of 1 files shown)


samuelgruetter:
LGTM

PR #151, opened 2015-03-25
Update to Rxjava 1.0.8 & add onBackpressureDropDo to ExperimentalAPIs

zsxwing:


samuelgruetter:
In RxJava, both `onBackpressureDrop` methods (the one with and the one without an action) have the same name, but in RxScala, you introduce the ugly name `onBackpressureDropWithDropAction`. I guess that's because otherwise the implicit conversion to `ExperimentalAPI` doesn't kick in? Or would it not even work if it was directly in `Observable`?

zsxwing:
It's because `overload` and `implicit conversion` cannot work together. Better name? Or we should consider another approach to add experimental APIs?

zsxwing:
@samuelgruetter how about I just update RxJava? I would like to release a new version to test the build scripts.

samuelgruetter:
As you wish ;-) Or I'd also be fine with renaming `onBackpressureDropWithDropAction` into something better (but what...?)

zsxwing:
onBackpressureDropWith or onBackpressureDropWithAction?

jbripley:
onBackpressureDropDo would be somewhat analogues with the doOn*(Action) methods?

zsxwing:
> onBackpressureDropDo would be somewhat analogues with the doOn*(Action) methods?

I think onBackpressureDropDo is better. @samuelgruetter thoughts?

samuelgruetter:
+1 for onBackpressureDropDo

zsxwing:
Updated the method name.

samuelgruetter:
LGTM

PR #155, opened 2015-04-10
Use the right accounts for sonatype

rspieldenner:
I wiped out all the existing secret: there were 5 of them so you might need to add one back.


zsxwing:
The rest one is `PGP_ENCRYPTION_PASSWORD`. We don't use it any more. Thank you! It works now.

PR #157, opened 2015-05-06
Update the out-of-date example in "Notification" doc

elegoff:
After an update on OnCompleted (https://github.com/elegoff/RxScala/commit/ebff7ccd0819538e586332b1f5577d2c03a2857c) the example does not work if used without a small change for pattern matching

### DIFF
--- a/src/main/scala/rx/lang/scala/Notification.scala
+++ b/src/main/scala/rx/lang/scala/Notification.scala
@@ -71,7 +71,7 @@ sealed trait Notification[+T] {
  * Example:
  * {{{
  * import Notification._
- * Observable(1, 2, 3).materialize.subscribe(n => n match {
+ * Observable.just(1, 2, 3).materialize.subscribe(n => n match {
  *   case OnNext(v)     => println("Got value " + v)
  *   case OnCompleted => println("Completed")
  *   case OnError(err)  => println("Error: " + err.getMessage)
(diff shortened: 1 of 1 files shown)


src/main/scala/rx/lang/scala/Notification.scala
```diff
@@ -73,7 +73,7 @@ sealed trait Notification[+T] {
  * import Notification._
  * Observable(1, 2, 3).materialize.subscribe(n => n match {
  *   case OnNext(v)     => println("Got value " + v)
- *   case OnCompleted() => println("Completed")
+ *   case OnCompleted => println("Completed")
```
zsxwing, line 76: Good catch.

src/main/scala/rx/lang/scala/Notification.scala
```diff
@@ -73,7 +73,7 @@ sealed trait Notification[+T] {
  * import Notification._
  * Observable(1, 2, 3).materialize.subscribe(n => n match {
```
zsxwing: Could you also change `Observable(1, 2, 3)` to `Observable.just(1, 2, 3)`? `Observable(1, 2, 3)` does not work now.

elegoff: Hi,
As per your request , I did the following  :
https://github.com/ReactiveX/RxScala/compare/0.x...elegoff:0.x

Thanks

Eric

## 

Eric Le Goff
http://fr.linkedin.com/in/elegoff
@elegoff https://twitter.com/elegoff

zsxwing:
LGTM. Merged. Thank you!

PR #159, opened 2015-05-07
Correct documentation for Observable.merge

### LINKED ISSUE #158: scaladoc for Observable.merge incorrect
The scaladoc for the Observable.merge states:
`returns an Observable that emits items from this and that until this or that emits onError or onComplete.`

From what I've observed, it behaves like this for errors, but not completion.  The output of merge continues until the all merged Observables complete.

For example, the following code outputs '1 3', but according to the docs should output only '1' (as the first observable will complete after 1 second as it only has 1 element to emit)

```
val ob1 = Observable.just(1).delay(1 second)
val ob2 = Observable.empty.delay(2 seconds)
val ob3 = Observable.just(3).delay(3 seconds)

ob1.merge(ob2).merge(ob3).doOnNext(println(_)).subscribe()
```

If in agreement that the scaladoc is incorrect for this method, I can submit a pull request to fix.

Originally discussed on https://class.coursera.org/reactive-002/forum/thread?thread_id=698

lukestephenson:
Fixes #158

### DIFF
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -2279,7 +2279,7 @@ trait Observable[+T]
    * @param that
    *            an Observable to be merged
    * @return an Observable that emits items from `this` and `that` until 
-   *            `this` or `that` emits `onError` or `onComplete`.
+   *            `this` or `that` emits `onError` or both Observables emit `onCompleted`.
    */
   def merge[U >: T](that: Observable[U]): Observable[U] = {
     val thisJava: rx.Observable[_ <: U] = this.asJavaObservable


src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -2279,7 +2279,7 @@ trait Observable[+T]
    * @param that
    *            an Observable to be merged
    * @return an Observable that emits items from `this` and `that` until 
-   *            `this` or `that` emits `onError` or `onComplete`.
+   *            `this` or `that` emits `onError` or both observables emit `onComplete`.
```
zsxwing: nit: observables => Observables

zsxwing: "onComplete" => "onCompleted"

lukestephenson:
@zsxwing Thanks for the feedback.  I've made those changes.

samuelgruetter:
LGTM, thanks

zsxwing:
Merged. Thank you!

PR #163, opened 2015-05-17
Update to Scala 2.10.5 / 2.11.6

jbripley:


zsxwing:
@jbripley could you elaborate your thought? You just want to update to the latest version, or you found some backward compatibility issue in 2.10.5 or 2.11.6? I think it's the former but just want to confirm with you.

jbripley:
It's the former. Just want to bump to latest version so we don't fall to far behind, especially with 2.11. 

Don't know of any current issues that these two upgrades fixes and going by our unit tests, doesn't cause any regressions either.

zsxwing:
@jbripley  Thanks.

PR #168, opened 2015-05-30
Move examples

hermanbanken:
This pull request moves the examples so that they no longer have access to private methods in the rx.lang.scala package. These examples being in the same package already made `TestScheduler.scala` not work out of the box.

As a novice RxScala-test-writer :sweat_smile: I did not understand at first that some JavaConversion was going on, as I missed that the mock class was mocking the Java variant of Observer. 
- Move examples
- Fix small missing java conversion in scaladoc

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,7 +31,7 @@ o1.switchIfEmpty(alternate).foreach(println)
 o2.switchIfEmpty(alternate).foreach(println)
 ```
 
-See more examples in [ExperimentalAPIExamples](https://github.com/ReactiveX/RxScala/blob/0.x/examples/src/test/scala/rx/lang/scala/examples/ExperimentalAPIExamples.scala)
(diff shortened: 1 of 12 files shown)


samuelgruetter:
Cool, thanks :-)

But now all pages linking to https://github.com/ReactiveX/RxScala/blob/0.x/examples/src/test/scala/rx/lang/scala/examples/RxScalaDemo.scala will have to be updated. For instance the [README](https://github.com/ReactiveX/RxScala/blob/0.x/README.md) or the [RxScala website](http://reactivex.io/rxscala/) and probably many more...

I'd try to update as many as possible of them, but we'll probably miss some, so I'd suggest that you add a commit to this PR which adds an `RxScalaDemo.scala` file where it was before, containing only a comment with the new link.

hermanbanken:
@samuelgruetter I included a file at the original RxScalaDemo.scala's location and updated some Markdown files as well.

samuelgruetter:
LGTM, thanks :+1:

zsxwing:
Merged. Thank you.

PR #170, opened 2015-06-06
Add TestSubscriber

zsxwing:
This PR adds TestSubscriber to RxScala. 

There are two differences between RxJava TestSubscriber and RxScala TestSubscriber.
- RxScala uses a singleton `Notification.OnCompleted`, so instead of adding `getOnCompletedEvents` to RxScala, I add  `getOnCompletedEventsCount` that just return the number of `Notification.OnCompleted`s.
- RxJava has two similar methods: `assertReceivedOnNext(List<T> items)` and `assertValues(T... values)`. Since they are almost same exception the parameter type, I only add `assertValues(T... values)` to RxScala.

Question:
- Do we need to add `TestObserver`? Is there a case that `TestObserver` is more useful than `TestSubscriber`?
- Better name for `getOnCompletedEventsCount`?

### DIFF
--- a/examples/src/test/scala/examples/TestSubscriberExample.scala
+++ b/examples/src/test/scala/examples/TestSubscriberExample.scala
@@ -1,0 +1,146 @@
+/**
+ * Copyright 2015 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


zsxwing:
cc @samuelgruetter

samuelgruetter:
> - Do we need to add `TestObserver`? Is there a case that `TestObserver` is more useful than `TestSubscriber`?

I can't think of any such case. Let's not add it until some users request it.

> - Better name for `getOnCompletedEventsCount`?

I would not add this method, nor `getOnCompletedEvents`. Instead, one should use `assertCompleted` and `assertNotCompleted`.

zsxwing:
> I would not add this method, nor getOnCompletedEvents. Instead, one should use assertCompleted and assertNotCompleted.

Agree. Removed it.

samuelgruetter:
LGTM

Issue #39, opened 2014-10-08
Adding -Xfatal-warnings


samuelgruetter:
Warnings by the Scala compiler often reveal problems with the code, and there's the `-Xfatal-warnings` option which makes the build fail if there are warnings. Currently, there are no warnings in this project, so we could add the `-Xfatal-warnings` option without problems, and be sure that further PRs don't introduce any warnings.

I tried to add the option to `build.sbt`, but it only applies to the root project but not to the examples project. Apparently `scalacOptions in ThisBuild` means "really only this project without sub-projects". But that would mean that the other options also don't apply to the examples project, which would be bad... /cc @jmhofer @zsxwing

headinthebox:
Let's hope we don't have any warnings.

samuelgruetter:
`build.sbt` is not set in stone ;-) So if at some point we get a false warning that we can't silence, we can just remove the `-Xfatal-warnings` option.

zsxwing:
There are some warnings when compiling with Scala 2.11.

```
[warn] /Users/zsx/workspace-scala/RxScala/src/test/scala/rx/lang/scala/CompletenessTest.scala:268: method paramss in trait MethodSymbolApi is deprecated: Use `paramLists` instead
[warn]       val paramListStrs = for (paramList <- m.paramss) yield {
[warn]                                               ^
[warn] /Users/zsx/workspace-scala/RxScala/src/test/scala/rx/lang/scala/CompletenessTest.scala:281: method declarations in class TypeApi is deprecated: Use `decls` instead
[warn]     methodMembersToMethodStrings(tp.declarations.filter {
[warn]                                     ^
[warn] /Users/zsx/workspace-scala/RxScala/src/test/scala/rx/lang/scala/CompletenessTest.scala:296: method companionSymbol in trait SymbolApi is deprecated: Use `companion` instead, but beware of possible changes in behavior
[warn]       getPublicInstanceMethods(tp.typeSymbol.companionSymbol.typeSignature)
[warn]                                              ^
[warn] /Users/zsx/workspace-scala/RxScala/src/test/scala/rx/lang/scala/CompletenessTest.scala:319: method companionSymbol in trait SymbolApi is deprecated: Use `companion` instead, but beware of possible changes in behavior
[warn]                    typeOf[rx.Observable[_]].typeSymbol.companionSymbol.typeSignature)
[warn]                                                        ^
[warn] four warnings found
```

jbripley:
[...]

PR #174, opened 2015-08-23
Refactor CompletenessTest to support adding completeness tests for other classes

zsxwing:
`Single` was added to RxJava recently. It would be better if we could also test its completeness when adding it to RxScala.

This PR refactors CompletenessTest to support testing completeness of other classes easily, including:
- Add `CompletenessKit` as a basic trait for classes that we want to test completeness. Most of codes are from the old `CompletenessTest.scala`.
- Add ObservableCompletenessKit.scala, TestSchedulerCompletenessKit.scala, TestSubscriberCompletenessKit.scala and BlockingObservableCompletenessKit.scala to demonstrate how to use `CompletenessKit`.
- Move the codes used to generate the completeness table to CompletenessTables.scala and change it to support to output tables for all `CompletenessKit`s. Run `sbt 'test:run rx.lang.scala.completeness.CompletenessTest'` to see these tables.
- Move all classes for completeness to the `completeness` package.


build.sbt
```diff
@@ -16,6 +16,8 @@ scalaVersion in ThisBuild := "2.11.6"
 
 crossScalaVersions in ThisBuild := Seq("2.10.5", "2.11.6")
 
+parallelExecution in Test := false
```
zsxwing, line 19: Disabled parallel execution because I encounter some weird exception without this line, e.g.,

```
[info] - checkScalaMethodPresenceVerbose *** FAILED ***
[info]   java.lang.ClassNotFoundException: o .lang.scala.Observ
[info]   at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[info]   at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
[info]   at java.lang.Class.forName0(Native Method)
[info]   at java.lang.Class.forName(Class.java:348)
[info]   at scala.reflect.runtime.JavaMirrors$JavaMirror.javaClass(JavaMirrors.scala:500)
[info]   at scala.reflect.runtime.SymbolLoaders$TopClassCompleter.complete(SymbolLoaders.scala:32)
[info]   at scala.reflect.internal.Symbols$Symbol.info(Symbols.scala:1231)
[info]   at scala.reflect.internal.Types$TypeRef.thisInfo(Types.scala:2407)
[info]   at scala.reflect.internal.Types$TypeRef.baseClasses(Types.scala:2412)
```

I guess some reflection APIs are not thread-safe (or a bug?). So if running multiple `CompletenessKit` at the same time, it will crash.

However, since RxScala's unit tests are pretty fast, disabling it hurts nothing.

zsxwing:
BTW, SBT [0.13.8](https://github.com/sbt/sbt/releases/tag/v0.13.8) supports cross-version for Scala sources:

> Cross-version support for Scala sources
> When crossPaths setting is set to true (it is true by default), sbt 0.13.8 will include
> src/main/scala-<scalaBinaryVersion>/ to the Compile compilation in addition to
> src/main/scala. For example, it will include src/main/scala-2.11/ for Scala 2.11.5, and
> src/main/scala-2.9.3 for Scala 2.9.3.

We can upgrade SBT to `0.13.8` and move these completeness tests to the `scala-2.11` folder. So that we can fix the deprecation warnings and enable `-Xfatal-warnings`. It's safe to run completeness tests only in Scala 2.11.

src/test/scala/rx/lang/scala/CompletenessTest.scala
```diff
@@ -1,494 +0,0 @@
...
-  @Ignore // because spams output
-  @Test def printDefaultMethodCorrespondence(): Unit = {
-    println("\nDefault Method Correspondence")
-    println(  "-----------------------------\n")
-    val c = SortedMap(defaultMethodCorrespondence.toSeq : _*)
-    val len = c.keys.map(_.length).max + 2
-    for ((javaM, scalaM) <- c) {
-      println(s"""      %-${len}s -> %s,""".format("\"" + javaM + "\"", "\"" + scalaM + "\""))
-    }
-  }
```
samuelgruetter, line 366: Several methods (including this one) are deleted here but not added to anywhere. I would like to keep them, because they might be useful for maintenance in the future.

src/test/scala/rx/lang/scala/completeness/BlockingObservableCompletenessKit.scala
```diff
@@ -0,0 +1,40 @@
...
+  override protected def correspondenceChanges = Map(
+    "first(Func1[_ >: T, Boolean])" -> "[use `Observable.filter(p).toBlocking.head`]",
+    "firstOrDefault(T)" -> "headOrElse(=> U)",
+    "firstOrDefault(T, Func1[_ >: T, Boolean])" -> "[use `Observable.filter(p).toBlocking.headOrElse(=> U)`]",
+    "forEach(Action1[_ >: T])" -> "foreach(T => Unit)",
+    "from(Observable[_ <: T])" -> "[use `Observable.toBlocking`]",
+    "last(Func1[_ >: T, Boolean])" -> "[use `Observable.filter(p).toBlocking.last`]",
+    "lastOrDefault(T)" -> "lastOrElse(=> U)",
+    "lastOrDefault(T, Func1[_ >: T, Boolean])" -> "[use `Observable.filter(p).toBlocking.lastOrElse(=> U)`]",
+    "mostRecent(T)" -> "mostRecent(U)",
```
samuelgruetter, line 34: We could also change the signature of RxScala's `mostRecent` and get rid of this line...

src/test/scala/rx/lang/scala/completeness/TestSchedulerCompletenessKit.scala
```diff
@@ -0,0 +1,26 @@
...
+package rx.lang.scala.completeness
+
+import scala.reflect.runtime.universe.typeOf
+
+class TestSchedulerCompletenessKit extends CompletenessKit {
+  override val rxJavaType = typeOf[rx.schedulers.TestScheduler]
+
+  override val rxScalaType = typeOf[rx.lang.scala.schedulers.TestScheduler]
+
+  override protected def correspondenceChanges: Map[String, String] = Map("triggerActions()" -> "triggerActions()")
```
samuelgruetter, line 26: It took me a while to figure out while this line is needed: It's because `CompletenessKit.javaMethodSignatureToScala` removes `()` from Java methods by default (and that's good). But I'd find it useful to add this explanation in the source code here, because it's quite puzzling ;-)

src/test/scala/rx/lang/scala/completeness/TestSubscriberCompletenessKit.scala
```diff
@@ -0,0 +1,41 @@
...
+    "assertNotCompleted()" -> "assertNotCompleted()",
+    "assertReceivedOnNext(List[T])" -> "assertValues(T*)",
+    "assertTerminalEvent()" -> "assertTerminalEvent()",
+    "assertUnsubscribed()" -> "assertUnsubscribed()",
+    "awaitTerminalEvent()" -> "awaitTerminalEvent()",
+    "getOnCompletedEvents()" -> "assertCompleted()",
+    "onCompleted()" -> "onCompleted()",
+    "onStart()" -> "onStart()",
+    "unsubscribe()" -> "unsubscribe()"
+  )
```
samuelgruetter, line 35: You could add a boolean flag to `CompletenessKit` which controls whether `CompletenessKit.javaMethodSignatureToScala` removes `()` from Java methods or not, and override this flag in the subclasses so that we don't need all these "changes" which aren't changes.

samuelgruetter:
sbt 0.13.8 + [`-Xfatal-warnings`](https://github.com/ReactiveX/RxScala/issues/39) sounds nice, let's do that [until we get some "unfixable" warnings ;-)]

samuelgruetter:
Nice re-design :+1:
I only had a few nitpicky remarks to add ;-)

zsxwing:
@samuelgruetter just sent more commits to address your comments.

zsxwing:
Merged your fix.

samuelgruetter:
LGTM

PR #178, opened 2015-10-05
Eliminated two possible false positives from BlockingObservableTest

illusioni:


zsxwing:
Could you clarify why you need to change them, such as provide an example of `false positives`?

illusioni:
Hi Shixiong Zhu,
The existing test is wrong because it does not test all that's specified.
It only works because the component under test (CUT) behaves well.
In the case of the "toFuture function throwing rather than producing a failing future" the existing test would still succeed (that's the false positive).
My version of the test would fail - as a test should when the CUT does not comply with the spec.
Hope that info helps. :-)
Big fan of Rx by now...
Klaus
On Oct 6, 2015 17:00, Shixiong Zhu notifications@github.com wrote:Could you clarify why you need to change them, such as provide an example of false positives?

src/test/scala/rx/lang/scala/observables/BlockingObservableTest.scala
```diff
@@ -161,15 +161,23 @@ class BlockingObservableTest extends JUnitSuite {
...
 
-  @Test(expected = classOf[IllegalArgumentException])
+  @Test
   def testToFutureWithMultipleItems() {
     val o = Observable.just(1, 2)
-    Await.result(o.toBlocking.toFuture, 10 seconds)
+    val future = o.toBlocking.toFuture //if this were to throw the original test would wrongly succeed
+    Await.result(future.failed, 10 seconds) match {
+      case t:IllegalArgumentException => //this is what we expect
+      case _ => fail("expected the future to fail with a NoSuchElementException")
```
zsxwing: nit: NoSuchElementException -> IllegalArgumentException

zsxwing:
@illusioni Thanks for your clarification. I will merge this PR once you fix the typo.

illusioni:
I tried hard, @zsxwing, but I can't find a typo in my changes.
Please let me know where you see an error

illusioni:
Aaah. Is this what you meant?

zsxwing:
@illusioni Exactly. Thanks! Merged.

PR #182, opened 2015-11-22
Upgrade to RxJava 1.0.17

zsxwing:


src/test/scala-2.11/rx/lang/scala/completeness/ObservableCompletenessKit.scala
```diff
@@ -76,6 +76,8 @@ class ObservableCompletenessKit extends CompletenessKit {
     "doOnUnsubscribe(Action0)" -> "doOnUnsubscribe(=> Unit)",
     "doOnTerminate(Action0)" -> "doOnTerminate(=> Unit)",
     "elementAtOrDefault(Int, T)" -> "elementAtOrDefault(Int, U)",
+    "extend(Func1[_ >: OnSubscribe[T], _ <: R])" -> "[TODO]",
+    "fromCallable(Callable[_ <: T])" -> "[TODO]",
```
zsxwing: Any better name in Scala instead of `fromCallable`?

zsxwing:
CC @samuelgruetter

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: scala
 jdk:
 - oraclejdk7
 script:
-- sbt +test +examples/test
+- sbt +test +examples/test +doc
```
zsxwing, line 5: Also add `doc` to the build script so that we can test it for PRs

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -132,7 +132,7 @@ trait Observable[+T]
    * $noDefaultScheduler
    *
    * @return $subscribeAllReturn
-   * @throws OnErrorNotImplementedException if the [[Observable]] tries to call `onError`
+   * @throws rx.exceptions.OnErrorNotImplementedException if the [[Observable]] tries to call `onError`
```
zsxwing, line 135: Use the full path to make Scaladoc generate the links

zsxwing:
@samuelgruetter could you take a look?

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -332,6 +332,143 @@ trait Observable[+T]
...
+   * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
+   */
+  @Experimental
+  def concatMapEager[R](f: T => Observable[R], capacityHint: Int): Observable[R] = {
+    toScalaObservable[R](asJavaObservable.concatMapEager[R](new Func1[T, rx.Observable[_ <: R]] {
+      def call(t1: T): rx.Observable[_ <: R] = {
+        f(t1).asJavaObservable
+      }
+    }, capacityHint))
+  }
```
samuelgruetter, line 469: In `def flatMap[R](maxConcurrent: Int, f: T => Observable[R]): Observable[R]`, we have the integer as the first argument, and the function as the second argument. Any reason why to do it differently here? And do you know why they put the function first in RxJava?

build.sbt
```diff
@@ -19,7 +19,64 @@ crossScalaVersions in ThisBuild := Seq("2.10.5", "2.11.6")
...
+doc in Compile <<= (doc in Compile) map { target: File =>
+  (target ** "*.html").get.filter(hasJavadocLink).foreach { f =>
+    val newContent: String = allExternalJavadocLinks.foldLeft(IO.read(f)) {
+      case (oldContent: String, javadocURL: String) =>
+        javadocLinkRegex(javadocURL).replaceAllIn(oldContent, fixJavaLinks)
+    }
+    IO.write(f, newContent)
+  }
+  target
+}
```
samuelgruetter, line 82: not exactly what I'd call a maintainable build setup... ;-) but as long as it works, I'm ok with it

src/main/scala/rx/lang/scala/Observable.scala
zsxwing, line 469: Just forgot we moved `maxConcurrent` as the first parameter in #166. Will updated it.

> And do you know why they put the function first in RxJava?

I guess it just followed the signature in Rx .Net (https://msdn.microsoft.com/en-us/library/hh229923(v=vs.103).aspx)

src/test/scala-2.11/rx/lang/scala/completeness/ObservableCompletenessKit.scala
samuelgruetter: Can't we just use `Observable.defer(Observable.just(expensiveComputation))`? If yes, can you add an example for this to `RxScalaDemo`?

zsxwing: It works as well. Agree that we don't need to add a new method for it.

zsxwing:
Updated

samuelgruetter:
LGTM, thanks a lot @zsxwing :)

PR #189, opened 2016-02-20
Upgrade to RxJava 1.1.1 and add missing methods

zsxwing:
- Upgrade to RxJava 1.1.1
- Add `doAfterTerminate` and deprecate `finallyDo`
- Add `cacheWithInitialCapacity` and deprecate `cache(Int)`
- Add `ConnectableObservable.autoConnect`


zsxwing:
cc @samuelgruetter

samuelgruetter:
just noticed that observeOn with delayError flag and cacheWithInitialCapacity are not in RxScalaDemo, but maybe that's fine because they're not super interesting... ;-)
LGTM

zsxwing:
Thanks, @samuelgruetter I will add examples later.

PR #191, opened 2016-04-13
Scala 2.12 support

zsxwing:
- Add Scala 2.12-M4 support
- Move APICoverage.scala to scala-2.11 as it doesn't support Java 8. Only running it in Scala 2.11 is enough.
- Update the Travis CI script to run Scala 2.12 build.


samuelgruetter:
cool, LGTM

PR #193, opened 2016-05-25
Reimplement tail, to and toMultimap

zsxwing:
Use existing RxJava operators to implement `tail`, `to` and `toMultimap` rather than using `lift`.


samuelgruetter:
Nice :smiley: 
And I see that all changed methods have tests which still pass  :+1: 
LGTM

PR #194, opened 2016-06-06
Bump to RxJava 1.1.5

zsxwing:
- Bump to RxJava 1.1.5
- Add missing operators


samuelgruetter:
I see that more and more `xxxDelayError` methods are added, and the number of methods in the API becomes bigger and bigger. Currently there are 8 of them:

```
  def concatDelayError[U](implicit evidence: Observable[T] <:< Observable[Observable[U]]): Observable[U]
  def concatMapDelayError[R](f: T => Observable[R]): Observable[R]
  def switchMapDelayError[R](f: T => Observable[R]): Observable[R]
  def switchDelayError[U](implicit evidence: Observable[T] <:< Observable[Observable[U]]): Observable[U]
  def mergeDelayError[U >: T](that: Observable[U]): Observable[U]
  def flattenDelayError[U](implicit evidence: Observable[T] <:< Observable[Observable[U]]): Observable[U]
  def flattenDelayError[U](maxConcurrent: Int)(implicit evidence: Observable[T] <:< Observable[Observable[U]]): Observable[U]
  def combineLatestDelayError[T, R](sources: Iterable[Observable[T]])(combineFunction: Seq[T] => R): Observable[R]
```

Will there be even more later?
Do you think we could treat all of these in a more uniform way? Maybe by adding an `ErrorDelayingObservable` containing all these methods, and a `toErrorDelaying` method, in the same way as we have a `BlockingObservable`? Or maybe give these methods a boolean argument with default value: `delayError: Boolean = false`? Or something else? Or is this too much complication for what we gain? Also, do you know if this has been discussed in RxJava, and what was decided for RxJava?

zsxwing:
> Maybe by adding an ErrorDelayingObservable containing all these methods, and a toErrorDelaying method, in the same way as we have a BlockingObservable

It will become too verbose for RxJava considering Java doesn't have the extension methods.

> Or maybe give these methods a boolean argument with default value: delayError: Boolean = false?

Java doesn't have the default value.

However, both them are supported in Scala.

For the default values, I remember when we use overload, default value, currying and implicit together, the compiler will be crazy and not be able to resolve the correct method. So I prefer the extension method idea.

samuelgruetter:
I'm a bit confused, why you are mentioning extension methods? Do you mean extension methods as defined [here](https://coderwall.com/p/4clu3a/extension-methods-in-scala)? The approach used in BlockingObservable does not require extension methods at all.

zsxwing:
Yep. `toErrorDelaying` is too verbose so I suggest to use the extension methods.

samuelgruetter:
Ok, but what if we call it `.delayError` instead of `toErrorDelaying`? Then, instead of `myObservable.concatMapDelayError(myFunc)`, we'd write `myObservable.delayError.concatMap(myFunc)`, which has exactly the same amount of letters. What do you think about this?

zsxwing:
Sounds great. I will update the PR.

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -5160,12 +5235,63 @@ object Observable {
...
+   *
+   * @tparam T the common base type of source values
+   * @tparam R the result type
+   * @param sources the [[scala.collection.Iterable Iterable]] of source [[Observable]]s
+   * @param combineFunction the aggregation function used to combine the items emitted by the source [Observable]]s
+   * @return an [[Observable]] that emits items that are the result of combining the items emitted by the source
+   *         [[Observable]]s by means of the given aggregation function
+   * @see <a href="http://reactivex.io/documentation/operators/combinelatest.html">ReactiveX operators documentation: CombineLatest</a>
+   */
+  def combineLatestDelayError[T, R](sources: Iterable[Observable[T]])(combineFunction: Seq[T] => R): Observable[R] = {
```
zsxwing, line 5289: `combineLatestDelayError` is still in `Observable` object.

src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -2725,37 +2804,6 @@ trait Observable[+T]
...
-   *
-   * $noDefaultScheduler
-   *
-   * @param maxConcurrent the maximum number of [[Observable]]s that may be subscribed to concurrently
-   * @return an [[Observable]] that emits all of the items emitted by the [[Observable]]s emitted by `this`
-   * @see <a href="http://reactivex.io/documentation/operators/merge.html">ReactiveX operators documentation: Merge</a>
-   * @since (if this graduates from Experimental/Beta to supported, replace this parenthetical with the release number)
-   */
-  @Experimental
-  def flattenDelayError[U](maxConcurrent: Int)(implicit evidence: Observable[T] <:< Observable[Observable[U]]): Observable[U] = {
```
zsxwing, line 2750: Since this is experimental, I just moved it into ErrorDelayingObservable.

zsxwing:
@samuelgruetter updated.

src/test/scala-2.11/rx/lang/scala/completeness/ObservableCompletenessKit.scala
```diff
@@ -200,8 +202,8 @@ class ObservableCompletenessKit extends CompletenessKit {
     "combineLatest(Iterable[_ <: Observable[_ <: T]], FuncN[_ <: R])" -> "combineLatest(Iterable[Observable[T]])(Seq[T] => R)",
     "combineLatestDelayError(Iterable[_ <: Observable[_ <: T]], FuncN[_ <: R])" -> "combineLatest(Iterable[Observable[T]])(Seq[T] => R)",
     "concat(Observable[_ <: Observable[_ <: T]])" -> "concat(<:<[Observable[T], Observable[Observable[U]]])",
-    "concatDelayError(Observable[_ <: Observable[_ <: T]])" -> "concatDelayError(<:<[Observable[T], Observable[Observable[U]]])",
-    "concatDelayError(Iterable[_ <: Observable[_ <: T]])" -> "[use `iter.toObservable.concatDelayError`]",
+    "concatDelayError(Observable[_ <: Observable[_ <: T]])" -> "[use `delayError.concatDelay(<:<[Observable[T], Observable[Observable[U]]])`]",
+    "concatDelayError(Iterable[_ <: Observable[_ <: T]])" -> "[use `iter.toObservable.delayError.concatDelay`]",
```
samuelgruetter: should be `concat` instead of `concatDelay`, I guess?

src/main/scala/rx/lang/scala/Observable.scala
samuelgruetter, line 5289: Yeah, I guess `Observable.combineLatestDelayError` is better than  `ErrorDelayingObservable.combineLatest` (unless there will be many more static `xxxDelayError` methods, maybe)

src/test/scala-2.11/rx/lang/scala/completeness/ObservableCompletenessKit.scala
zsxwing: > should be concat instead of concatDelay, I guess?

Fixed

samuelgruetter:
LGTM

PR #196, opened 2016-06-15
Deprecate Observable.create

zsxwing:
Let's deprecate `Observable.create` now. We can add it back when the coursera course is open again.


samuelgruetter:
linking to https://github.com/ReactiveX/RxJava/issues/1340#issuecomment-225391215 for future reference

samuelgruetter:
LGTM

PR #199, opened 2016-06-29
Fix the wrong return type of `take` overload with scheduler

jking-roar:
fixes signature of method with erroneous Unit return type


zsxwing:
Good catch. Thanks! Could you also add a simple example in https://github.com/ReactiveX/RxScala/blob/0.x/examples/src/test/scala/examples/RxScalaDemo.scala to test the signature?

zsxwing:
Thanks! @jking-roar

PR #203, opened 2016-08-21
Add assertValuesAndClear and something test features.

imcharsi:
However, this commit's test is failed because there are various changes between 1.1.6 and 1.1.9, and I cannot decide about these problems myself.

Something Problems:
  In `rx.Scheduler`, `when` method was appended.
  In `rx.Observable`, 24 methods are appended.

And, While I put `assertValuesAndClear`'s test into `SubscriberTests`, but you should move to appropriate place for this test or remove if this test is not needed.

You should consider this commit as reference or even it would be better to create your own from scratch.

**Please Don't Merge this PR as is!!!**


zsxwing:
@imcharsi thanks for your contribution. I merged your commit into #204.

PR #207, opened 2016-09-16
RxScala <-> RxJava converters

rvanheest:
Hi, based on my question on [StackOverflow](http://stackoverflow.com/questions/39505292/convert-rxjava-observable-to-rxscala-observable) and the answer/invite by @samuelgruetter I got to work and created a `JavaConverters` class that is similar to the `rx.lang.scala.JavaConversions` class but does the conversion between RxJava and RxScala by using the `asScala` method (and `asJava` vise versa). The implementation is similar to the one in the Scala SDK (`scala.collections.JavaConverters`) and offers a way to do a more _explicit_ implicit conversion.

**Usage:** although the use is quite obvious, I included an example of the intended use below.
Given a function that return a RxJava `Observable` (most likely coming from a Java based API), you transform it to a RxScala `Observable` using `asScala`:

``` scala
import rx.lang.scala.JavaConverters._

object DemoJ2S extends App {

  def getJavaObservable: rx.Observable[Int] = rx.Observable.just(1, 2, 3)

  getJavaObservable.asScala.subscribe(i => println(i))
}
```

Similarly, you convert a RxScala `Observable` to an RxJava `Observale` using `asJava`:

``` scala
object DemoS2J extends App {

    def getScalaObservable: Observable[Int] = Observable.just(1, 2, 3)

    getScalaObservable.asJava.subscribe(new Action1[Int] {
        override def call(i: Int): Unit = println(i)
    })
}
```

Besides `Observable` I included similar conversions for all types that were already in `rx.lang.scala.JavaConversions` (`Observer`, `Scheduler`, `Worker`, etc.).

### DIFF
--- a/examples/src/test/scala/examples/JavaConvertersDemo.scala
+++ b/examples/src/test/scala/examples/JavaConvertersDemo.scala
@@ -1,0 +1,84 @@
+package examples
+
+import java.util.concurrent.CountDownLatch
+
+import org.junit.{Ignore, Test}
+import org.scalatest.junit.JUnitSuite
+import rx.functions.Action1
+import rx.lang.scala.JavaConverters._
+import rx.lang.scala.{Observable, Observer, Subscriber, Subscription}
+
+import scala.language.postfixOps
+
+@Ignore // Since this doesn't do automatic testing, don't increase build time unnecessarily
+class JavaConvertersDemo extends JUnitSuite {
+
+  @Test
+  def javaObservableToScalaObservableConverter() = {
+
+    // given a RxJava Observable (for example from some third party library)...
+    def getJavaObservableFromSomewhere: rx.Observable[Int] =
+      rx.Observable.just(1, 2, 3)
+
+    // it is possible to transform this into a RxScala Observable using `asScala`
+    // after that you can use all Scala style operators
+    getJavaObservableFromSomewhere.asScala
+      .map(2 *)
+      .filter(_ % 4 == 0)
+      .subscribe(println, _.printStackTrace(), () => println("done"))
+  }
+
+  @Test
+  def scalaObservableToJavaObservableConverter() = {
+
+    // given a function that takes a RxJava Observable as its input and returns a RxJava Subscription ...
+    def useJavaObservableSomewhere(obs: rx.Observable[_ <: Int]): rx.Subscription =
+      obs.subscribe(new Action1[Int] { override def call(i: Int): Unit = println(i) })
+
+    // you can give a RxScala Observable as input and call `asJava`on it...
+    val javaSubscription: rx.Subscription = useJavaObservableSomewhere(Observable.just(1, 2, 3).asJava)
+
+    // after which you get a RxJava Subscription back, which can be used again in a RxScala setting using `asScalaSubscription`...
+    val scalaSubscription: Subscription = javaSubscription.asScalaSubscription
+
+    // and convert it back to a RxJava Subscription using `asJavaSubscription`
+    val javaSubscriptionAgain: rx.Subscription = scalaSubscription.asJavaSubscription
+  }
+
+  @Test
+  def schedulerConverting() = {
+    // the next line is not part of the actual example, but is just to prevent early termination due to scheduling
+    val latch = new CountDownLatch(1)
+
+
+    // given a function that return a particular RxJava-specific scheduler (for example the
+    // JavaFxScheduler from RxJavaFx (https://github.com/ReactiveX/RxJavaFX))...
+    def getJavaSpecificScheduler: rx.Scheduler =
+      rx.schedulers.Schedulers.computation()
+
+    Observable.just(1, 2, 3)
+      .doOnNext(_ => println(Thread.currentThread().getName)) // should print "main"
+      .observeOn(getJavaSpecificScheduler.asScala)
+      .doOnNext(_ => println(Thread.currentThread().getName)) // should print the name of the scheduler as its side effect
+      .subscribe(println, e => { e.printStackTrace(); latch.countDown() }, () => { println("done"); latch.countDown() })
+
+
+    latch.await()
+  }
+
+  def observableCreate(): Unit = {
+    Observable((subscriber: Subscriber[Int]) => {
+      // As a subscriber is both a Subscriber, Observer and Subscription, you need some way to distinguish between them
+      // Observer uses the regular `asJava` and `asScala` operators;
+      // Subscriber uses `asJavaSubscriber` and `asScalaSubscriber`;
+      // Subscription uses the `asJavaSubscription` and `asScalaSubscription` instead.
+      val jObserver: rx.Observer[_ >: Int] = subscriber.asJava
+      val jSubscriber: rx.Subscriber[_ >: Int] = subscriber.asJavaSubscriber
+      val jSubscription: rx.Subscription = subscriber.asJavaSubscription
+
+      val sObserver: Observer[_ >: Int] = jObserver.asScala
+      val sSubscriber: Subscriber[_ >: Int] = jSubscriber.asScalaSubscriber
+      val sSubscription: Subscription = jSubscription.asScalaSubscription
+    })
+  }
+}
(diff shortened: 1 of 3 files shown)


src/main/scala/rx/lang/scala/JavaConverters.scala
```diff
@@ -0,0 +1,122 @@
...
+object JavaConverters extends DecorateAsJava with DecorateAsScala
+
+private[scala] trait Decorators {
+	class AsJava[A](op: => A) {
+		def asJava: A = op
+	}
+
+	class AsJavaSubscription(s: Subscription) {
+		def asJavaSubscription: rx.Subscription = s.asJavaSubscription
+	}
```
samuelgruetter: Why don't you use `asJava` for subscriptions as well?

src/main/scala/rx/lang/scala/JavaConverters.scala
```diff
@@ -0,0 +1,122 @@
...
+	* Provides conversion functions `asJava` and `asScala` to convert
+	* between RxScala types and RxJava types.
+	*
+	* Example:
+	* {{{
+	* import rx.lang.scala.JavaConverters._
+	* val javaObs = Observable.just(1, 2, 3).asJava
+	* val scalaObs = javaObs.asScala
+	* }}}
+	*/
```
samuelgruetter: indentation should be one space before the `*`

src/main/scala/rx/lang/scala/JavaConverters.scala
```diff
@@ -0,0 +1,122 @@
...
+object JavaConverters extends DecorateAsJava with DecorateAsScala
+
+private[scala] trait Decorators {
```
samuelgruetter, line 34: (sorry for nitpicking) Code indentation in this project is 2 spaces, could you please adhere to this?

samuelgruetter (changes requested):
I like the idea, and it seems to be independent from the [JavaConversions](https://github.com/ReactiveX/RxScala/blob/0.x/src/main/scala/rx/lang/scala/JavaConversions.scala), so the two approaches could coexist.

I'd like to hear also @zsxwing's opinion.

Before this can be merged, there should be a test-and-example file for it, similar to [RxScalaDemo](https://github.com/ReactiveX/RxScala/blob/0.x/examples/src/test/scala/examples/RxScalaDemo.scala).
The goal is not to test the _functionality_ (that's trivial), but it's to test the _signatures_: That they can be used the in the intended way, without overloading problems, conflicting implicits, or implicit conversions not kicking in, etc... There's plenty of things which could go wrong here, especially at a later point when the library evolves. And moreover, the tests can also serve as a guide with usage examples.

I'd suggest that you create a new test file, in the same directory as the RxScalaDemo, where you show by some examples how the `asScala`/`asJava` extension methods can be used. And ideally, you'd also explain that there's a second approach in `JavaConversions`.

rvanheest:
_in this response the RxJava variants of the types are prefixed with a `j`_

Let me reply to your questions:

**Could I add some examples similar to `RxScalaDemo`?** - I actually had some examples but didn't commit them as they were not real tests. Did not see `RxScalaDemo` before. I added them 55b816b. This was also helpful for me, as I found some type level bugs with it!

**spaces instead of tabs** - fixed that

**Why don't I use `asJava` for `Subscription`?** - I did, but there are some type level problems with this, because of the relation between `Subscription`, `Subscriber` and `Observer`. For example, if you would have a `Subscriber[T]` somewhere (say in an `Observable.apply`), then calling `asJava` gives a `jSubscriber[_ >: T]` (welcome to covariance and contravariance!) but downcasting to `jObserver[_ >: T]` wouldn't work. To fix this the only solution is to give the extension method another name. Same thing (as I discovered in 55b816b) holds for `jSubscription`! You have to give it other names. As I have experienced that the most use of `asJava` and `asScala` is in `jObservable` to `Observable` and `jObserver` to `Observer`, I figured I use `asJava` and `asScala` for the `Observer` variant and `asXSubscriber` and `asXSubscription` for `Subscriber` and `Subscription`... I was not able to find any other solution. If you know a better one, feel free to suggest this.

Hope this clears up any of your questions.

examples/src/test/scala/examples/JavaConvertersDemo.scala
```diff
@@ -0,0 +1,84 @@
...
+      val jObserver: rx.Observer[_ >: Int] = subscriber.asJava
+      val jSubscriber: rx.Subscriber[_ >: Int] = subscriber.asJavaSubscriber
+      val jSubscription: rx.Subscription = subscriber.asJavaSubscription
+
+      val sObserver: Observer[_ >: Int] = jObserver.asScala
+      val sSubscriber: Subscriber[_ >: Int] = jSubscriber.asScalaSubscriber
+      val sSubscription: Subscription = jSubscription.asScalaSubscription
+    })
+  }
+}
```
samuelgruetter, line 84: This is an excellent demo :smiley:

samuelgruetter:
I tried to understand why we can't use `asScala`/`asJava` everywhere, so I quickly changed all `asScalaXxx` to `asScala` and also for Java, see https://github.com/ReactiveX/RxScala/pull/208.
(Aside: Please also have a look at commit https://github.com/ReactiveX/RxScala/pull/208/commits/eb6db125a5aeb97fd909fb7addf2df9aef911465: Always give an explicit return type for public methods, otherwise bad things can happen, see the commit message).
If you try to compile it, a `scala.reflect.internal.FatalError` occurs, you can see it in the [Travis CI log](https://travis-ci.org/ReactiveX/RxScala/jobs/161177714).
That is, we just found a bug in the Scala compiler :grimacing: 
Did you get this as well?
If you have time, could you please check if this also happens with the latest Scala compiler, and if yes, try to isolate and report it?
Finding a scalac bug with RxScala actually happened to me before, and the scalac maintainers could show me a simple work-around, so let's hope they'll know such a work-around for this one as well :wink:

rvanheest:
Yup, I got a similar error yesterday in the [last example](https://github.com/ReactiveX/RxScala/pull/207/files#diff-21b9149f7bff852c85a70ea3c58bde1bR69) as well (I run Scala 2.11.8). Asked about it on https://gitter.im/scala/scala and got a response from @SethTisue (see below).

If I find time in the next couple of days, I'll have a look at this compiler bug. Let's keep in touch on this!

In the meantime I'll fix the `Unit` types you briefly mentioned...

---

https://files.gitter.im/scala/scala/6nER/blob

screen shot 2016-09-20 at 06 32 56

screen shot 2016-09-20 at 06 28 15

zsxwing:
Since `scala.collection.JavaConversions` could cause a lot of problems implicitly and is deprecated in Scala 2.12.0, I suggest that we also deprecate `rx.lang.scala.JavaConversions`.

rvanheest:
@zsxwing I did not know that, but indeed read this in the [scaladoc](http://www.scala-lang.org/files/archive/api/2.12.0-RC1/scala/collection/JavaConversions$.html). Maybe this PR could provide a replacement for this. Although I really think that deprecating `rx.lang.scala.JavaConversions` should be done in a separate PR.

@samuelgruetter I tried to come up with a better solution than using separate keywords for `Subscriber` and `Subscription`, but kept walking into this same compiler error. Maybe you could propose a better solution if you know any?

Other than that, I consider this PR as a nice alternative for `rx.lang.scala.JavaConversions`.

zsxwing:
> Although I really think that deprecating rx.lang.scala.JavaConversions should be done in a separate PR.

Agreed.

samuelgruetter:
I isolated the scalac bug:

``` scala
import scala.language.implicitConversions

object Bug extends App {

  class JObserver[T]
  class JSubscriber[T] extends JObserver[T]

  class Foo

  implicit def convertSubscriber[T](s: JSubscriber[_ >: T]): Foo = ???

  implicit def convertObserver[T](s: JObserver[_ >: T]): Foo = ???

  val jSubscriber: JSubscriber[_ >: Int] = ???

  val foo: Foo = jSubscriber // compiler crashes

}
```

I think it's [SI-5559](https://issues.scala-lang.org/browse/SI-5559), and it seems that there is no simple work-around...

So I guess we have to live with the more verbose `asScalaXxx` methods, and I'd say LGTM, ready to merge. Thanks @rvanheest!

rvanheest:
My pleasure, mate! You guys are doing great stuff and I'm happy to contribute.

samuelgruetter (approved):

Issue #209, opened 2016-09-26
Conversions from Scala's Try and Option to Observable


rvanheest:
Would it make any sense to have conversion methods from `scala.util.Try` and `scala.Option` to `rx.lang.scala.Observable`? I mean, there is a conversion from `Iterable` to `Observable` in `rx.lang.scala.ObservableExtensions`, which I view as the Scala equivalent of RxJava's `Observable.from`.

Java does not really have type level notions of `Try` and `Option` (the latter only in Java 8), so it makes sense that RxJava does not have them. Scala, however, _does_ have these types and in my view they should have a similar conversion method as `Iterable`.

The implementations would of course be very simple:
- `Option`
  - `Some(v)` -> `OnNext(v)` followed by `OnCompleted`
  - `None` -> `OnCompleted`
- `Try`
  - `Success(v)` -> `OnNext(v)` followed by `OnCompleted`
  - `Failure(e)` -> `OnError(e)`

LMKWYT! I'd be more than happy to do another PR!

zsxwing:
It's not necessary to add an Option overload. It's already supported. See

``` Scala
scala> Observable.from(Some(1)).foreach(println)
1
```

Try is missing. So PR is welcome!

rvanheest:
Agree kind of agree on the `Option`, although it seems that you can't do `Some(1).toObservable` with that. I think that is because the compiler has to do 2 implicit conversions for that: the `Iterable[T]` to `Option[T]` and the `toObservable` conversion. So that's why I suggested to do the `Option` part as well.

Regarding the `Try`, I have made a similar thing as the `ObservableExtensions` in the package object, so the syntax would be `Try(...).toObservable`. But now that I'm thinking about it, maybe you would like to have the `Observable.from(Try(...))` alternative as well... I'll add this last bit and do a PR. Then we'll see what we can use and what can be left out!

rvanheest:
Closing, as this was solved in #210

PR #210, opened 2016-09-26
Try & Option converters

rvanheest:
Refer to issue #209 for the start of the discussion.

This PR adds a `.toObservable` converter for `scala.util.Try` and `scala.Option` as well as a `Observable.from` wrapper for `scala.util.Try`.

### DIFF
--- a/src/main/scala/rx/lang/scala/Observable.scala
+++ b/src/main/scala/rx/lang/scala/Observable.scala
@@ -20,6 +20,7 @@ import rx.annotations.{Beta, Experimental}
 import rx.exceptions.OnErrorNotImplementedException
 import rx.functions.FuncN
 import rx.lang.scala.observables.{ConnectableObservable, ErrorDelayingObservable}
+
 import scala.concurrent.duration
 import java.util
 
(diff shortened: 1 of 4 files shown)


src/main/scala/rx/lang/scala/Observable.scala
```diff
@@ -5042,6 +5043,24 @@ object Observable {
   }
 
   /**
+   * Converts a `Try` into an `Observable`.
+   *
```
rvanheest, line 5047: Not sure who makes them, but we may or may not want a marble diagram here, just as with most of the other operators/factory methods. If so, I'd be happy to put a `` in here!

samuelgruetter: There's some info on the marble diagrams [here](https://github.com/ReactiveX/reactivex.github.io#marble-diagrams), but I don't think we need one here, converting a `Try` into an `Observable` is quite a straightforward operation I guess :wink:

samuelgruetter (approved):
cool, LGTM

zsxwing:
👍 Thanks, @rvanheest

PR #211, opened 2016-10-01
Bump to RxJava 1.2.0

zsxwing:


src/test/scala/rx/lang/scala/ObservableTest.scala
```diff
@@ -400,7 +400,7 @@ class ObservableTests extends JUnitSuite {
   @Test
   def testToMap() {
     val expectedMap1 = (0 to 100).map(i => (i % 2, i)).toMap
-    val m1 = (0 to 100).toObservable.toMap(_ % 2)
+    val m1 = (0 to 100).toObservable.toMap[Int](_ % 2)
```
zsxwing, line 403: Seems Scala 2.12.0-RC1 has a regression. Just created https://issues.scala-lang.org/browse/SI-9943

samuelgruetter: wow... curious to see whether they'll fix it :wink:

samuelgruetter (approved):
LGTM

Issue #100, opened 2014-12-19
Using RxJava @Beta/@Experimental features in RxScala


jbripley:
What is the current thinking about using features/methods marked as @Beta or @Experimental in RxJava and exposing them in RxScala?

For example, I was thinking of doing a PR that would make AbstractOnSubscribe available in RxScala with a more Scala idiomatic API. Is that something that would be accepted or should we wait until the feature is marked non-beta/experimental before exposing them in RxScala.

Downside of exposing beta/experimental methods in RxScala is that we tie each RxScala version tighter to a specific RxJava release, with the upside being of making the beta/experimental features available earlier with a Scala interface.

headinthebox:
Why don't we add them as extension methods (perhaps using the same markers) so you can import them separately.

zsxwing:
> Downside of exposing beta/experimental methods in RxScala is that we tie each RxScala version tighter to a specific RxJava release, with the upside being of making the beta/experimental features available earlier with a Scala interface.

I think we can document two tables for the API compatibility, one for stable APIs, the other for `@Beta/@Experimental` APIs. If someone uses any `@Beta/@Experimental` API, they need to use the exactly same version of RxJava.

zsxwing:
@headinthebox for the experimental methods of RxJava Observable, I added them as extension methods in RxScala #146. But for the experimental classes, such as `AbstractOnSubscribe`, I think we can also use RxJava `@Beta/@Experimental`. Thoughts?

zsxwing:
Closed. Resolved in 0.25.0

PR #213, opened 2016-10-16
Add UnicastSubject

ruippeixotog:
This pull request adds a wrapper for JavaRx `UnicastSubject`. Although the `Subject` is experimental, I would like to see a Scala wrapper for it as it's very useful in one of my projects.

I read #100, but I'm still not sure of how I should mark classes as experimental. I opted to use the annotation that is used in Java, but please tell me if you want it another way and I'll change it.

### DIFF
--- a/src/main/scala/rx/lang/scala/subjects/UnicastSubject.scala
+++ b/src/main/scala/rx/lang/scala/subjects/UnicastSubject.scala
@@ -19,25 +19,29 @@ import rx.annotations.Experimental
 import rx.lang.scala.Subject
 
 /**
-  * A `Subject` variant which buffers events until a single `Subscriber` arrives and replays them to it
-  * and potentially switches to direct delivery once the `Subscriber` caught up and requested an unlimited
-  * amount. In this case, the buffered values are no longer retained. If the `Subscriber`
-  * requests a limited amount, queueing is involved and only those values are retained which
-  * weren't requested by the `Subscriber` at that time.
+  * $experimental A `Subject` variant which buffers events until a single `Subscriber` arrives and replays
+  * them to it and potentially switches to direct delivery once the `Subscriber` caught up and requested an
+  * unlimited amount. In this case, the buffered values are no longer retained. If the `Subscriber` requests
+  * a limited amount, queueing is involved and only those values are retained which weren't requested by the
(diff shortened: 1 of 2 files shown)


src/main/scala/rx/lang/scala/subjects/UnicastSubject.scala
```diff
@@ -0,0 +1,52 @@
...
+import rx.lang.scala.Subject
+
+/**
+  * A `Subject` variant which buffers events until a single `Subscriber` arrives and replays them to it
+  * and potentially switches to direct delivery once the `Subscriber` caught up and requested an unlimited
+  * amount. In this case, the buffered values are no longer retained. If the `Subscriber`
+  * requests a limited amount, queueing is involved and only those values are retained which
+  * weren't requested by the `Subscriber` at that time.
+  */
+@Experimental
```
samuelgruetter, line 31: Putting the `@Experimental` annotation here is fine, but in addition, you could also add a red badge in the scaladoc comment, as in [ErrorDelayingObservable](https://github.com/ReactiveX/RxScala/blob/0.x/src/main/scala/rx/lang/scala/observables/ErrorDelayingObservable.scala).

samuelgruetter (changes requested):
Thanks for contributing :smiley: 
I have just one minor scaladoc nitpick (see below), otherwise I like it :+1:

src/main/scala/rx/lang/scala/subjects/UnicastSubject.scala
ruippeixotog: Done, it looks fine!

zsxwing: @ruippeixotog you forgot to add `$experimental` to the class doc.

ruippeixotog: Done. I also added `@Experimental` to the methods in order to be consistent with what was done in `ErrorDelayingObservable`.

samuelgruetter (approved):

samuelgruetter (approved):
LGTM

zsxwing:
👍

zsxwing:
Thanks @ruippeixotog

Issue #219, opened 2016-11-24
(A)SyncOnSubscribe


dhoepelman:
Hi,

are there any plans on implementing RxJava 1.2's `AsyncOnSubscribe` and `SyncOnSubscribe`?
I could contribute an implementation, but am unsure about the preferred signature.

```
// 1: Java style, thin wrapper but un-idiomatic
Observable(new SyncOnSubscribe[State, T](
    override def generateState: () => State
    // User must ensure Observer gets exactly one call to onNext, onCompleted or onError like in RxJava
    override def next: (Observer[_ <: T], State) => State
    override def onUnsubscribe: State => ()
))
```

```
// 2: User is responsible for correctly calling observer, similar to RxJava
object SyncOnSubscribe {
  def apply[State, T](
    generateState: () => State,
    next: (State, Observer[_ <: T]) => State,
    onUnsubscribe: State => () = _ => ()
  )
}

object AsyncOnSubscribe {
  def apply[State, T](
    generateState: () => State,
    next: (State, Long, Observer[Observable[_ <: T]]) => State,
    onUnsubscribe: State => () = _ => ()
  )
}
```

```
// 3: User returns one of 3 valid values from an ADT
// Seems cleanest, safest and most convenient for the user, but worth the overhead and divergence from RxJava?

// 3a: our own ADT
sealed trait ObserverValue[T] // Could use a better name
object ObserverValue {
    case class Next[T](value: T) extends ObserverValue
    case object Completed extends ObserverValue
    case object Error(e: Throwable) extends ObserverValue
}

// 3b: We could also implement this using standard scala types. The type would not be named in the real implementation
type ObserverValue[T] = Try[Option[T]]
// Failure(e) => onError(e), Success(None) => onCompleted, Success(Some(t)) => onNext(t)

// 3c: We could also use an observable, and only pick it's first element (or completed/error)
type ObserverValue[T] = Observable[T]

object SyncOnSubscribe {
  def apply[State, T](
    generateState: () => State,
    next: State => (ObserverValue[T], State),
    onUnsubscribe: State => () = _ => ()
  )
}

object AsyncOnSubscribe {
  def apply[State, T](
    generateState: () => State,
    next: (State, Long) => (Observable[T], State),
    onUnsubscribe: State => () = _ => ()
  )
}
```

I'm leaning towards 3b, but see good arguments for 2 and all 3's.

samuelgruetter:
[...]

Issue #175, opened 2015-09-08
doOnCompleted never executes


gkhom:
Using RxScala 0.25.0, the following example never calls the lambda that is provided to `doOnCompleted`:

``` Scala
    Observable.empty
      .doOnCompleted(() => println("DONE 1"))
      .subscribe(x=>x, e=>e, () => println("DONE 2"))
```

The same holds for this example:

``` Scala
    Observable[Unit](observer => {
      observer.onCompleted()
    }).doOnCompleted(() => println("DONE 1"))
      .subscribe(x=>x, e=>e, () => println("DONE 2"))
```

Both examples only print `DONE 2`.

Either I'm overlooking something obvious, or this is a bug.

samuelgruetter:
You're overlooking something, but it's not obvious at all ;-)

The signature of `doOnCompleted` is:

``` scala
def doOnCompleted(onCompleted: => Unit): Observable[T]
```

That is, it takes a _by-name_ argument. So you have to use it as follows:

``` scala
Observable.empty
        .doOnCompleted(println("DONE 1"))
        .subscribe(x=>x, e=>(), () => println("DONE 2"))
```

or, to convince yourself that the action passed to `doOnCompleted` is not already run when constructing the observable, you can write

``` scala
    println("line1")
    val obs1 = Observable.empty.doOnCompleted(println("DONE 1"))
    println("line2")
    obs1.subscribe(x=>x, e=>(), () => println("DONE 2"))
```

which prints

```
line1
line2
DONE 1
DONE 2
```

What you were passing to `doOnCompleted` is a _0-arity function_, i.e. an expression of type `() => Unit`, and by discarding the value of an expression, Scala can turn any expression into `Unit`, so that's why it compiled.

Originally, `doOnCompleted` took `() => Unit`, which I find more intuitive and safer. But then, it was "simplified" into `=> Unit` in https://github.com/ReactiveX/RxJava/pull/1345. 
So maybe we should revert that change?

gkhom:
@samuelgruetter Thank you for the clarification, I completely overlooked the by-name parameter. It's quite confusing though since `subscribe` does take the onCompleted parameter as a 0-arity function but `doOnCompleted` does it with a by-name parameter. I think it's better to stay consistent and just use the same parameter type everywhere. If this is up for debate my preference would be `() => Unit` over `=> Unit` since it's more explicit IMO.

samuelgruetter:
[...]

PR #220, opened 2016-11-25
Implement SyncOnSubscribe and AsyncOnSubscribe

dhoepelman:
See #219 for some discussion.

I implemented both option 2 (as `createX`) and 3 (as `applyX`), as 2 has some major annoyances when used from scala: type interferencing can't determine the `T` and the user requires at least 2 statements because you need to call the observer (you can't use a single expression).
The effects of this can be seen next to each other in the tests.

However, this causes duplicate very-similar functionality. If we want to avoid this I suggest we drop the java style `createX` methods and rename the `applyX` methods to `createX`.

### DIFF
--- a/src/main/scala/rx/lang/scala/Notification.scala
+++ b/src/main/scala/rx/lang/scala/Notification.scala
@@ -63,6 +63,16 @@ sealed trait Notification[+T] {
   }
 
   def apply(observer: Observer[T]): Unit = accept(observer)
+
+  // TODO: Should this be public or private to rx.lang.scala?
+  def map[U](f: T => U): Notification[U] = {
+    this match {
+      case Notification.OnNext(value) => Notification.OnNext(f(value))
+      // TODO: Or do we want to cast here? Notification.OnError[T] is not a Notification[U]
+      case Notification.OnError(error) => Notification.OnError(error)
+      case Notification.OnCompleted => Notification.OnCompleted
+    }
+  }
 }
 
 /**
--- a/src/main/scala/rx/lang/scala/Notification.scala
+++ b/src/main/scala/rx/lang/scala/Notification.scala
@@ -64,11 +64,9 @@ sealed trait Notification[+T] {
 
   def apply(observer: Observer[T]): Unit = accept(observer)
 
-  // TODO: Should this be public or private to rx.lang.scala?
   def map[U](f: T => U): Notification[U] = {
     this match {
       case Notification.OnNext(value) => Notification.OnNext(f(value))
-      // TODO: Or do we want to cast here? Notification.OnError[T] is not a Notification[U]
       case Notification.OnError(error) => Notification.OnError(error)
       case Notification.OnCompleted => Notification.OnCompleted
     }
--- a/src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
+++ b/src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
@@ -1,40 +1,18 @@
 package rx.lang.scala.observables
 
 import rx.annotations.Experimental
-import rx.lang.scala.{Notification, Observable, Observer}
+import rx.lang.scala.{Notification, Observable}
 
 /**
-  * An utility class to create [[Observable]]s that start acting when subscribed to and responds
+  * An utility class to create `Observable`s that start acting when subscribed to and responds
   * correctly to back pressure requests from subscribers.
   *
   * Semantics:
   * * `generator` is called to provide an initial state on each new subscription
   * * `next` is called with the last state and a `requested` amount of items to provide a new state
-  *     and an [[Observable]] that (potentially asynchronously) emits up to `requested` items.
+  *     and an `Observable` that (potentially asynchronously) emits up to `requested` items.
   * * `onUnsubscribe` is called with the state provides by the last next when the observer unsubscribes
-  *
-  * @tparam S the type of the user-define state
-  * @tparam T the type items that this `AsyncOnSubscribe` will emit.
   */
-class AsyncOnSubscribe[S,T](val generator: () => S,
-                            val next: (S, Long) => (Notification[Observable[T]], S),
-                            val onUnsubscribe: S => Unit) { self =>
-  import rx.lang.scala.JavaConversions._
-
-  private[scala] val asJavaAsyncOnSubscribe = new rx.observables.AsyncOnSubscribe[S,T] {
-    override def generateState(): S = generator()
-    override def next(state: S, requested: Long, observer: rx.Observer[rx.Observable[_ <: T]]): S =
-      self.next(state, requested) match {
-        case (notification, nextState) =>
-          toJavaNotification(notification.map(toJavaObservable)).accept(observer)
-          nextState
-      }
-    override def onUnsubscribe(state: S): Unit = self.onUnsubscribe(state)
-  }
-
-  def toObservable: Observable[T] = toScalaObservable[T](rx.Observable.create(asJavaAsyncOnSubscribe))
-}
-
 object AsyncOnSubscribe {
 
   /**
@@ -55,8 +33,23 @@ object AsyncOnSubscribe {
     * @param onUnsubscribe clean up behavior
     */
   @Experimental
-  def stateful[S, T](generator: () => S)(next: (S, Long) => (Notification[Observable[T]], S), onUnsubscribe: S => Unit = (_:S) => ()): AsyncOnSubscribe[S,T] =
-    new AsyncOnSubscribe[S, T](generator, next, onUnsubscribe)
+  def stateful[S, T](generator: () => S)(next: (S, Long) => (Notification[Observable[T]], S), onUnsubscribe: S => Unit = (_:S) => ()): AsyncOnSubscribe[S,T] = {
+    // The anonymous class shadows these names
+    val nextF = next
+    val onUnsubscribeF = onUnsubscribe
+
+    new rx.observables.AsyncOnSubscribe[S,T] {
+      import rx.lang.scala.JavaConversions._
+      override def generateState(): S = generator()
+      override def next(state: S, requested: Long, observer: rx.Observer[rx.Observable[_ <: T]]): S =
+        nextF(state, requested) match {
+          case (notification, nextState) =>
+            toJavaNotification(notification.map(toJavaObservable)).accept(observer)
+            nextState
+        }
+      override def onUnsubscribe(state: S): Unit = onUnsubscribeF(state)
+    }
+  }
 
   /**
     * Generates a [[AsyncOnSubscribe]] which does not generate a new state in `next`
--- a/src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
+++ b/src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
@@ -11,7 +11,7 @@ import rx.lang.scala.{Notification, Observable}
   * * `generator` is called to provide an initial state on each new subscription
   * * `next` is called with the last state and a `requested` amount of items to provide a new state
   *     and an `Observable` that (potentially asynchronously) emits up to `requested` items.
-  * * `onUnsubscribe` is called with the state provides by the last next when the observer unsubscribes
+  * * `onUnsubscribe` is called with the state provided by the last `next` call when the observer unsubscribes
   */
 object AsyncOnSubscribe {
 
--- a/src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
+++ b/src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
@@ -4,117 +4,82 @@ import rx.annotations.Experimental
 import rx.lang.scala.{Notification, Observable, Observer}
 
 /**
-  * An utility class to create observables that start acting when subscribed to and respond correctly
-  * to back pressure requests from subscribers.
+  * An utility class to create [[Observable]]s that start acting when subscribed to and responds
+  * correctly to back pressure requests from subscribers.
+  *
+  * Semantics:
+  * * `generator` is called to provide an initial state on each new subscription
+  * * `next` is called with the last state and a `requested` amount of items to provide a new state
+  *     and an [[Observable]] that (potentially asynchronously) emits up to `requested` items.
+  * * `onUnsubscribe` is called with the state provides by the last next when the observer unsubscribes
+  *
+  * @tparam S the type of the user-define state
+  * @tparam T the type items that this `AsyncOnSubscribe` will emit.
   */
-object AsyncOnSubscribe {
-  /**
-    * Generates an `Observable` that synchronously calls the provided `next` function
-    * to generate data to downstream subscribers.
-    *
-    * @tparam T the type of the generated values
-    * @tparam S the type of the associated state with each Subscriber
-    * @param generator generates the initial state value
-    * @param next produces data to the downstream subscriber
-    * @param onUnsubscribe clean up behavior
-    * @return An Observable that emits data downstream in a protocol compatible with back-pressure.
-    * @see rx.observables.AsyncOnSubscribe.createStateful
-    */
-  @Experimental
-  def createStateful[S, T](generator: () => S)(next: (S, Long, Observer[Observable[_ <: T]]) => S, onUnsubscribe: S => Unit = (_:S) => ()): Observable[T] =
-    new AsyncOnSubscribeImpl[S, T](generator, next, onUnsubscribe).toObservable
+class AsyncOnSubscribe[S,T](val generator: () => S,
+                            val next: (S, Long) => (Notification[Observable[T]], S),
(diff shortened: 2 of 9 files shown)


samuelgruetter:
I didn't have time yet for a full review, but maybe let's first discuss some desirable properties of an (A)SyncOnSubscribe solution for RxScala. I have the following in mind:

1. Should be compatible with future deprecation of `Observable.create`, see https://github.com/ReactiveX/RxJava/pull/4253
2. Should be reasonably close to RxJava, so that documentation, blog posts, forums etc on RxJava are also applicable to RxScala
3. The "right way" of constructing Observables (i.e. through (A)SyncOnSubscribe instead of home-made usage of `Observable.create`) should be easily discoverable by browsing the method list of Observable. In RxJava, this was achieved by adding unnecessary overloads of `Observable.create`, see https://github.com/ReactiveX/RxJava/pull/3738
4. In Scala code, if I write `MyObject(args)` (which is desugared into `MyObject.apply(args)`), I should get back something of type `MyObject`.

This PR does well on 1) and 2), but not so well on 3) and 4). Regarding 4), I write `AsyncOnSubscribe(...)`, but I get back an `Observable`, that's very confusing.

I don't have a solution for 3) and 4) yet, but will keep thinking about it... @dhoepelman @zsxwing and others, please let me know what you think, and if you have any ideas.

dhoepelman:
4 is a good point, I'll fix that.

I would add a requirement **5**: "It should feel natural so use from a Scala context". 
`SyncOnSubscribe.createStateful(() => state)((state, obs: Observer[_ <: Value]) => {obs.onNext(value); modify(state)})`
fails that, while
`SyncOnSubscribe.applyStateful(() => state)(state => (Notification.OnNext(value), state))`
 doesn't or less so. It is somewhat at odds with 2.

For **3** We could add the two (currently misnomered) `(A)SyncOnSubscribe.apply` methods to the `Observable` object as `Observable.(a)syncOnSubscribe`.

We then have to think about whether we still want to provide the `createStateless`/`createSingleState` utility methods either on the `Observable` object, an `SyncOnSubscribe` object or leave them out altogether*. I think it's better leaving them all out together over creating six new methods on the `Observable` object. Providing the methods on a separate object should be okay, but hurts discoverability.

\* We could assume a Scala programmer is sufficiently familiar with similar state concepts through things like state monads, `Observable#scan` or `Traversable#foldLeft` to ad-hoc construct them

samuelgruetter:
I agree with 5).

A note regarding wildcards in type parameters: All RxScala types have declaration site variance annotations (`+` or `-`), e.g. `Observer[-T]` or `Observable[+T]`. So, when you write `Observable[T]`, it always means `Observable[_ <: T]`, so you don't need to write the wildcard. (This applies to your comment above as well as to some lines in the code of this PR).

Here's an idea for 3):

* Add two methods to Observable: `Observable.create(rx.lang.scala.observables.AsyncOnSubscribe)` and `Observable.create(rx.lang.scala.observables.SyncOnSubscribe)`
* Define `rx.lang.scala.AsyncOnSubscribe[S, -T]` as a type alias for `rx.observables.AsyncOnSubscribe[S, T]`, and same for Sync, and say in their scaladoc that instances should be created with the constructors provided in the Scala (A)SyncOnSubscribe

Like this, discoverability should be good: Users browse for Observable.create, then ask "how can I get an (A)SyncOnSubscribe", go to (A)SyncOnSubscribe, and find the constructors.
Note that since users don't interact with (A)SyncOnSubscribe directly, we don't need to wrap the Java object in a Scala object, but can just use a type alias.
And moreover, constructing an Observable in Scala will look the same as in Java: Just something like, for instance, `Observable.create(SyncOnSubscribe.createStateless(...))`.

WDYT?

src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
```diff
@@ -0,0 +1,120 @@
...
+import rx.annotations.Experimental
+import rx.lang.scala.{Notification, Observable, Observer}
+
+/**
+  * An utility class to create observables that start acting when subscribed to and respond correctly
+  * to back pressure requests from subscribers.
+  */
+object AsyncOnSubscribe {
+  /**
+    * Generates an `Observable` that synchronously calls the provided `next` function
```
rvanheest: shouldn't this be '**a**synchronously'? (also in other scaladoc sections in this object)

src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
```diff
@@ -0,0 +1,120 @@
...
+    * @tparam T the type of the generated values
+    * @tparam S the type of the associated state with each Subscriber
+    * @param generator generates the initial state value
+    * @param next produces data to the downstream subscriber
+    * @param onUnsubscribe clean up behavior
+    * @return An Observable that emits data downstream in a protocol compatible with back-pressure.
+    * @see rx.observables.AsyncOnSubscribe.createStateful
+    */
+  @Experimental
+  def createStateful[S, T](generator: () => S)(next: (S, Long, Observer[Observable[_ <: T]]) => S, onUnsubscribe: S => Unit = (_:S) => ()): Observable[T] =
```
rvanheest: Have you considered a call-by-name parameter for `generator`? Would look better in using it, I think... (also in other functions in this object)

src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
```diff
@@ -0,0 +1,120 @@
...
+    * @param next produces an observable with at most the requested amount of items for the stream
+    * @param onUnsubscribe clean up behavior
+    * @return An observable that emits data downstream in a protocol compatible with back-pressure.
+    */
+  @Experimental
+  def applyStateless[T](next: Long => Notification[Observable[_ <: T]], onUnsubscribe: () => Unit = () => ()) =
+    apply[Unit, T](() => ())((_,r) => (next(r), ()), _ => onUnsubscribe())
+
+  private[scala] class AsyncOnSubscribeImpl[S,T](val generatorF: () => S, val nextF: (S, Long, Observer[Observable[_ <: T]]) => S, val onUnsubscribeF: S => Unit) {
+    import rx.lang.scala.JavaConversions._
```
rvanheest: Since there were plans to deprecate `JavaConversions` (see [comment by @zsxwing](https://github.com/ReactiveX/RxScala/pull/207#issuecomment-249042192)), maybe you should pro-actively use `JavaConverters` instead. (not sure what the status of this plan currently is, though!)

rvanheest (commented):
Found some small things for you to consider.

src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
dhoepelman: I considered this. However the spec clearly states that `generator` will be called *each time* an observer subscribes. If we use call-by-name it will only be executed once very early into the program, this would be confusing if the generator causes a side effect.

It also doesn't fit with the intended use-case, e.g. new subscriber means set up a HTTP connection, for each `onNext` do a request on that connection.

It's only identical if the generator is a pure function which produces an immutable value.

Maybe worth an overload.

src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
dhoepelman: The `next` is still synchronously called (see [RxJava javadoc](https://github.com/ReactiveX/RxJava/blob/dc63e5350d08f08b4c1229df812f4a0a19f1083e/src/main/java/rx/observables/AsyncOnSubscribe.java#L159)). The async in the name comes from `next`'s resultant `Observable[T]`, which can asynchronously deliver up to `requested` items.

I'll see if I can improve the docs to make this more clear.

src/main/scala/rx/lang/scala/observables/AsyncOnSubscribe.scala
samuelgruetter: To make sure the `generator` is called each time an observer subscribes, you just have to put it into a `() => generator`, as we did in [doOnCompleted](https://github.com/ReactiveX/RxScala/blob/0.26.4/src/main/scala/rx/lang/scala/Observable.scala#L3891), so that's not a problem.
However, I'm not sure if we should use a by name argument here, because if people pass a `() => expr` instead of a by name argument, the errors (at compile time or sometimes even only at run time) can be confusing, see eg [here](https://github.com/ReactiveX/RxScala/issues/175).
And please be careful with overloads: If you want to use overloading, please add tests ensuring that the overloads don't conflict and don't harm type parameter inference, we've had bad experiences with overloads before...

dhoepelman: Ah thanks! Didn't know/think of that semantic.

Errors should generally less confusing, as generally the state won't be of type `Unit` so there's no automatic conversion from `() => Unit` to `Unit` as in the issue you linked. The stateless variants don't have a generator.

An advantage of `() => T` is that it makes it very explicit that it is a function (thus it will be executed, possible multiple times). I would not expect `bar()` to be run multiple times in `foo(bar())`, even if the signature is `foo(x: => T)`, knowledge of the inner working of `foo` is necessary to know that.

I'll let you/the maintainers decide.

samuelgruetter: Yes, it's not as bad as in the issue I linked, but there still might be confusing compile-time errors, and the readability issue you mention. So I'd prefer `() => T` over of by-name, even though this is a bit less Scala-idiomatic.

dhoepelman:
@samuelgruetter Missed your last comment, that sounds like a good solution. I'll change the PR to fit it somewhere in the following days and incorporate the other comments.

dhoepelman:
@samuelgruetter Updated the PR. Also added `map` to `Notification`, let me know if you agree.

I'm not entirely sure I got the variance right, you talked about `AsyncOnSubscribe[S, -T]` but shouldn't `T` be covariant similar to how it is in `Observable[+T]`?

Only way I could make it work was with:
`type SyncOnSubscribe[S, +T] = rx.observables.SyncOnSubscribe[S, _ <: T]`

src/main/scala/rx/lang/scala/Notification.scala
```diff
@@ -63,6 +63,16 @@ sealed trait Notification[+T] {
   }
 
   def apply(observer: Observer[T]): Unit = accept(observer)
+
+  // TODO: Should this be public or private to rx.lang.scala?
```
samuelgruetter: I think it's nice to have this public, and remove the TODO.

src/main/scala/rx/lang/scala/Notification.scala
```diff
@@ -63,6 +63,16 @@ sealed trait Notification[+T] {
...
+  def map[U](f: T => U): Notification[U] = {
+    this match {
+      case Notification.OnNext(value) => Notification.OnNext(f(value))
+      // TODO: Or do we want to cast here? Notification.OnError[T] is not a Notification[U]
```
samuelgruetter: No cast, because it would be a cast that cannot be checked by the JVM, which might create a compile time warning, which would have to be suppressed etc, to complicated... (Unless someone presents benchmarks suggesting that a cast performs better).

src/main/scala/rx/lang/scala/observables/SyncOnSubscribe.scala
```diff
@@ -0,0 +1,77 @@
...
+import rx.lang.scala.{Notification, Observable}
+
+/**
+  * An utility class to create `Observable`s that start acting when subscribed to and responds
+  * correctly to back pressure requests from subscribers.
+  *
+  * Semantics:
+  * * `generator` is called to provide an initial state on each new subscription
+  * * `next` is called with the last state to provide a data item and a new state for the next `next` call
+  * * `onUnsubscribe` is called with the state provides by the last next when the observer unsubscribes
```
samuelgruetter:
```
`onUnsubscribe` is called with the state provided by the last `next` call when the observer unsubscribes
```
(also in AsyncOnSubscribe)

src/main/scala/rx/lang/scala/observables/package.scala
```diff
@@ -24,4 +24,7 @@ package rx.lang.scala
  * in Scala, because we use a pair `(key, observable)` instead of `GroupedObservable`
  * and a pair `(startFunction, observable)` instead of `ConnectableObservable`. 
  */
-package object observables {}
+package object observables {
+  type SyncOnSubscribe[S, +T] = rx.observables.SyncOnSubscribe[S, _ <: T]
+  type AsyncOnSubscribe[S, +T] = rx.observables.AsyncOnSubscribe[S, _ <: T]
```
samuelgruetter, line 29: Yes, the variance is correct here. 
An `OnSubscribe[T]` is a function `Subscribe[T] => Unit`, and the left side of the `=>` is a negative position, but being the argument of `Subscriber` is also a negative position, and minus times minus = plus :wink:

src/test/scala-2.11/rx/lang/scala/completeness/ObservableCompletenessKit.scala
```diff
@@ -198,8 +198,8 @@ class ObservableCompletenessKit extends CompletenessKit {
     // manually added entries for Java static methods
     "amb(Iterable[_ <: Observable[_ <: T]])" -> "amb(Observable[T]*)",
     "create(OnSubscribe[T])" -> "apply(Subscriber[T] => Unit)",
-    "create(SyncOnSubscribe[S, T])" -> "[TODO]",
-    "create(AsyncOnSubscribe[S, T])" -> "[TODO]",
+    "create(SyncOnSubscribe[S, T])" -> "create(SyncOnSubscribe[S, T])",
+    "create(AsyncOnSubscribe[S, T])" -> "create(AsyncOnSubscribe[S, T])",
```
samuelgruetter: Can't you just delete these two lines? This list should only mention correspondences where Scala differs from Java, but here the signatures are the same.

src/test/scala/rx/lang/scala/observables/AsyncOnSubscribeTests.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.scalatest.junit.JUnitSuite
+import rx.lang.scala.observers.TestSubscriber
+import rx.lang.scala.{Notification, Observable}
+
+class AsyncOnSubscribeTests extends JUnitSuite {
+
+  @Test
+  def testStateful(): Unit = {
+    val last = 2000L
+    val sut = Observable.create(AsyncOnSubscribe(() => 0L)((count,demand) =>
```
samuelgruetter: Why `sut`? Could you please use a generally understandable name or abbreviation?

src/test/scala/rx/lang/scala/observables/AsyncOnSubscribeTests.scala
```diff
@@ -0,0 +1,65 @@
...
+      if(count > last)
+        (Notification.OnCompleted, count)
+      else {
+        val max = math.max(count + demand, last)
+        val next = Observable.from(count to max)
+        (Notification.OnNext(next), max+1)
+      }
+    ))
+    assertEquals((0L to last).toList, sut.toBlocking.toList)
+  }
```
samuelgruetter, line 24: This is a good example, could you please also add something like this to [`RxScalaDemo`](https://github.com/ReactiveX/RxScala/blob/0.x/examples/src/test/scala/examples/RxScalaDemo.scala)? Also, adding (A)SyncOnSubscribe requires some information in `RxScalaDemo` to be updated: `createExampleGood` is not good any more, but only "medium good" :wink: It respects unsubscription, but not backpressure, and using (A)SyncOnSubscribe would be better. Could you please review all examples related to `Observable.create`, and add some comments, and examples for (A)SyncOnSubscribe, so that this is all up to date?

samuelgruetter (changes requested):
I added some minor comments, but otherwise it's great :smiley:

src/test/scala/rx/lang/scala/observables/AsyncOnSubscribeTests.scala
dhoepelman: I thought this was a generally well known abbreviation for `subject under test` in unit tests, but apparently that's only the case in my local bubble. Will change.

dhoepelman:
@samuelgruetter I incorporated your comments and altered the RxDemo file

I rewrote all of the `create` examples to use (A)SyncOnSubscribe, this seems correct no? Since `rx.Observable.create(OnSubscribe)` is slated to be deprecated (at least in the public API) it would seem logical for RxScala to move that way too.

I'm not entirely sure about `createExampleWithBackpressure`, but it seems to me (A)SyncOnSubscribe provides a way more user-friendly way to achieve the same.

samuelgruetter (approved):
LGTM :+1: 

@zsxwing, could you please have a look at this as well before merging, since this is quite an important change, and also because you originally wrote `createExampleFromInputStream`.

zsxwing:
LGTM2.

zsxwing:
Thanks!

PR #222, opened 2016-12-23
Add FutureToObservable

dhoepelman:
Added `FutureToObservable` so all `Observable.from(T)` overloads have a matching `TToObservable`

Renamed `ObservableExtensions` to the more descriptive `IterableToObservable`, which also makes the name in line with `TryToObservable` and `OptionToObservable`.


samuelgruetter (approved):
Nice, LGTM

src/main/scala/rx/lang/scala/package.scala
```diff
@@ -24,11 +25,11 @@ import _root_.scala.util.{Try, Success, Failure}
...
 
-  /**
-   * Placeholder for extension methods into Observable[T] from other types
-   */
-  implicit class ObservableExtensions[T](val source: Iterable[T]) extends AnyVal {
-      def toObservable: Observable[T] = {  Observable.from(source) }
+  @deprecated("Use IterableToObservable", "0.26.5")
+  type ObservableExtensions[T] = IterableToObservable[T]
+
+  implicit class IterableToObservable[T](val iterable: Iterable[T]) extends AnyVal {
```
zsxwing: This actually will break binary compatibility. It's better to keep the class name and the parameter name because I think this implicit may be one of the most popular APIs.

dhoepelman: Ah oops, didn't think of binary comp

dhoepelman:
@zsxwing Removed the `ObservableExtensions` rename from the PR

zsxwing:
LGTM. Thanks!

PR #226, opened 2017-01-20
Implement Observable.fromEmitter

dhoepelman:
This was still missing in the Scala API.

I haven't encountered a lot of use for it within the scala ecosystem (since callback are uncommon), but it's useful when integrating non-scala JVM libraries.

I'll add documentation if this structure is found to be OK

### DIFF
--- a/src/main/scala/rx/lang/scala/Emitter.scala
+++ b/src/main/scala/rx/lang/scala/Emitter.scala
@@ -1,0 +1,44 @@
+package rx.lang.scala
+
+import rx.annotations.Experimental
+import rx.functions.Cancellable
+
+@Experimental
+trait Emitter[-T] extends Observer[T] {
+
+  private [scala] val asJavaEmitter: rx.Emitter[_ >: T] = new rx.Emitter[T] {
+    import JavaConverters.toScalaSubscription
+
+    override def requested(): Long = Emitter.this.requested
+    override def setCancellation(c: rx.functions.Cancellable): Unit = Emitter.this.setCancellation(() => c.cancel())
+    override def setSubscription(s: rx.Subscription): Unit = Emitter.this.setSubscription(s.asScalaSubscription)
+
+    override def onError(e: Throwable): Unit = Emitter.this.onError(e)
+    override def onCompleted(): Unit = Emitter.this.onCompleted()
+    override def onNext(t: T): Unit = Emitter.this.onNext(t)
+  }
+
+  def requested: Long
+
+  def setCancellation(onCancellation: () => Unit): Unit
+  def setSubscription(subscription: Subscription): Unit
+}
+
+object Emitter {
+  type BackpressureMode = rx.Emitter.BackpressureMode
+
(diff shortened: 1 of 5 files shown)


src/test/scala/rx/lang/scala/ConstructorTest.scala
```diff
@@ -32,4 +37,86 @@ class ConstructorTest extends JUnitSuite {
     assertEquals(List(1,2,3), xs)
 
   }
+
+  @Test def testFromEmitter(): Unit = {
```
dhoepelman, line 41: Not sure if `ConstructorTest.scala` isthe right place for these, can also move them to a new file

src/main/scala/rx/lang/scala/Emitter.scala
```diff
@@ -0,0 +1,41 @@
+package rx.lang.scala
+
+import rx.annotations.Experimental
+
+@Experimental
+trait Emitter[-T] extends Observer[T] {
```
dhoepelman: Now that I think about it, not sure this interface carries it's weight for a scala version.

It differs only 3 ways:

`setCancellation` takes a `() => Unit` instead of a `rx.functions.Cancellable` SAM type. This is super minor, as Scala can automatically convert the function type to a SAM type. No problem to use the java interface.

`setSubscription` takes a `rx.lang.scala.Subscription` instead of a `rx.Subscription`. This is annoying for users, as they would need to import JavaConvertors and always use `subscription.asJava`. However, it is not very difficult and `setSubscription` is a tad more advanced usage which people won't always use.

It inherits from `rx.lang.scala.Observer` instead of `rx.Observer`. I have now idea how common/valid the use case it to use a , but I suspect usually people will only use the interface directly in `Observable.fromEmitter(e => /* use e */)` so it isn't a problem

src/main/scala/rx/lang/scala/Emitter.scala
```diff
@@ -0,0 +1,44 @@
+package rx.lang.scala
+
+import rx.annotations.Experimental
+import rx.functions.Cancellable
+
+@Experimental
+trait Emitter[-T] extends Observer[T] {
```
dhoepelman, line 7: Now that I think about it, I'm not sure this interface carries it's weight for a scala version.

It's nearly identical, but differs in 3 ways:

setCancellation takes a () => Unit instead of a rx.functions.Cancellable SAM type. This is super minor for 2.12 users, as it can automatically convert the function type to the SAM type.

setSubscription takes a rx.lang.scala.Subscription instead of a rx.Subscription. This is annoying for users, as they would need to import JavaConvertors and always use subscription.asJava. However, it is not very difficult and setSubscription is a tad more advanced usage which people won't always use.

It inherits from rx.lang.scala.Observer instead of rx.Observer. I have now idea how common/valid the use case of passing it around as an Observer is , but I suspect usually people will only use the interface directly in Observable.fromEmitter(e => /* use e */) so it isn't a problem

thoughts?

PR #231, opened 2018-06-13
Implicit conversion to Scala Future for Observable

greatbalin:
Main idea of this change it to make it possible to simply move from Observable to Future when needed.


samuelgruetter:
* Why do you force users to specify a timeout? What if they want to wait indefinitely? I think the issue of converting from Observable to Future and the issue of timeouts are orthogonal and should not be mixed up.
* It seems you are kind of reinventing the wheel, couldn't you use something like `obs.toBuffer.toBlocking.toFuture`?

greatbalin:
Yes! you are right!

By some reason I totally forgot about `toBuffer` .

Issue #238, opened 2018-07-22
toSingle and toCompletable


Proximator:
Is there any plan to implement those two methods as in RxJava1?

dhoepelman:
No, see #186 and #187

PR #239, opened 2018-07-23
#238: Initial commit to implement Single and Completable in RxScala1

Proximator:
I would like to implement Single and Completable, I just started working on that
Could you please take a look to my code and check if this makes sense?
Thanks


samuelgruetter:
I would not invest any effort into building software on top of RxJava1, as RxJava1 [is end of life (EOL)](https://github.com/ReactiveX/RxJava/tree/1.x#end-of-life-notice).

dhoepelman:
Due to project EOL status, this will not be merged.

PR #243, opened 2019-08-06
Add Scala 2.13 support

dhoepelman:
Upgrade Scala 2.12 to 2.12.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,19 @@
+dist: xenial
 language: scala
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt
 matrix:
   include:
-    - jdk: openjdk7
+    - jdk: openjdk8
       scala: 2.10.7
-    - jdk: openjdk7
+    - jdk: openjdk8
       scala: 2.11.12
-    - jdk: oraclejdk8
-      scala: 2.12.6
+    - jdk: openjdk8
+      scala: 2.12.9
+    - jdk: openjdk8
+      scala: 2.13.0
 script:
 - sbt ++$TRAVIS_SCALA_VERSION test examples/test doc
 after_success:
--- a/build.sbt
+++ b/build.sbt
@@ -10,11 +10,16 @@ lazy val examples = project in file("examples") dependsOn (root % "test->test;co
   )
 )
 
-scalacOptions in ThisBuild := Seq("-feature", "-unchecked", "-deprecation", "-encoding", "utf8", "-Xfatal-warnings")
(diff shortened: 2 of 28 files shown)


.travis.yml
```diff
@@ -1,12 +1,15 @@
+dist: xenial
 language: scala
 matrix:
   include:
-    - jdk: openjdk7
+    - jdk: openjdk8
```
dhoepelman: Travis build with the previous JDK versions fails with:
```
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"

Ignoring license option: BCL -- using GPLv2+CE by default

install-jdk.sh 2019-07-17

Expected feature release number in range of 9 to 14, but got: 8

The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

build.sbt
```diff
@@ -10,11 +10,16 @@ lazy val examples = project in file("examples") dependsOn (root % "test->test;co
   )
 )
 
-scalacOptions in ThisBuild := Seq("-feature", "-unchecked", "-deprecation", "-encoding", "utf8", "-Xfatal-warnings")
+scalacOptions in ThisBuild := Seq("-feature", "-unchecked", "-deprecation", "-encoding", "utf8")
```
dhoepelman, line 13: Fatal warnings is infeasible in 2.13 since there are a lot of deprecation warnings, unsolvable while supporting 2.10-2.12.
scala-collection-compat does not support 2.10 and is incomplete, and would add a dependency.

dhoepelman:
@samuelgruetter or @zsxwing This library is not actively maintained anymore, but a 2.13 version seems useful for projects which still depend on it.

.travis.yml
```diff
@@ -1,12 +1,19 @@
+dist: xenial
```
dhoepelman, line 1: Had to do some Travis work since the build seems to have rotted

* Upgrade to a newer Ubuntu version with more JDK's installed
* Change openjdk7 and oraclejdk8 to openjdk8 since the Travis install JDK script now fails on JDKs <9 (openjdk8 is preinstalled)
* Downloading SBT takes very long and frequently times out (see builds of now-replaced commits), caching the `~/.sbt` directory should work around this (once it succeeds once it can reuse the cache) and is a good idea anyway since it increases the build speed. `~/.ivy2/cache` is there to cache dependencies for build speed.

